PR #3, opened 2017-07-07
Update repository with recent discussion and begin prototypes

starksm64:


api/src/main/java/org/eclipse/microprofile/jwt/JWTCallerPrincipal.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * A standard CallerPrincipal implementation that provides access to the JWT claims that are required by
+ * the microprofile token.
+ */
+public class JWTCallerPrincipal extends CallerPrincipal {
+    /** TODO: need to decide on a JWT parsing/verification library or create one */
+    private JsonWebToken jwt;
```
johnament: Each impl needs to decide that, not the spec.  The spec may require a base interface to be satisfied, but please avoid using Keycloak in the spec.

api/src/main/resources/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
johnament: We should only need a root LICENSE.

api/src/test/resources/jwk-content1.json
```diff
@@ -0,0 +1,38 @@
...
+  "exp": 1311281970,
+  "iat": 1311280970,
+  "auth_time": 1311280969,
+  "roles": [
+    "role-in-realm",
+    "user",
+    "manager"
+  ],
+  "groups": ["group1", "group2"],
+  "resource_access": {
```
johnament: Would a consumer of this specifically look for their service, and add proper groups/roles based on that?  How would a developer specify which if their service?

tck/pom.xml
```diff
@@ -49,36 +49,16 @@
...
+            <groupId>org.eclipse.microprofile.config</groupId>
+            <artifactId>microprofile-config-api</artifactId>
+            <version>${project.version}</version>
         </dependency>
-        
         <dependency>
-            <groupId>javax.enterprise</groupId>
-            <artifactId>cdi-api</artifactId>
-            <version>1.2</version>
+            <groupId>javax</groupId>
```
johnament: Can we avoid depending on the entirety of javaee?  I see it's done in a few places.

api/src/main/java/org/eclipse/microprofile/jwt/JWTCallerPrincipal.java
starksm64: I'll move this to an ri subproject as reference implementation as I want a default implementation to illustrate the concepts.

api/src/test/resources/jwk-content1.json
starksm64: I'm expecting that the security integration layer is what is handling the mapping from the target microservice based on it's CDI name and declared role requirements.

tck/pom.xml
starksm64: We can, I was just following what has been done elsewhere.

starksm64:
This was signed in an amend, but apparently that does not meet the signing requirements, so I'll do a new pull request with changes for the comments.

PR #4, opened 2017-07-19
Initial prototype for JWTCallerPrincipal and associated reference implementation

starksm64:
Signed-off-by: Scott Stark <starksm64@gmail.com>


starksm64:
Emily, ok, I'll move the ri module to its own repository.

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -17,131 +17,136 @@
...
-   		"nonce": "n-0S6_WzA2Mj",
    		"exp": 1311281970,
    		"iat": 1311280970,
    		"auth_time": 1311280969,
-      "realm_access": {
-          "roles": ["role-in-realm", "user", "manager"]
-      },
+      "roles": ["role-in-realm", "user", "manager"],
+      "groups": [],
       "resource_access": {
```
chunlongLiang: For "resource_access", how do you identify "resource server" or "service" for this spec? Do we want "resource server"  or "resource" identifier to be defined in this spec?  If we want "resource server" or "resource" claim in this interop spec,  we do need define this claim. Maybe we could refer to the draft of "Resource Indicators for OAuth 2.0", https://tools.ietf.org/html/draft-campbell-oauth-resource-indicators-02, seems want to standardize "resource" identification, although it is not finalized yet. If it takes longer time for us to define this "resource server" claim, could we focus on the simple JWT format without "resource server" for now, and we come back to address "resource server" in next version spec?

starksm64: So I have been working on prototypes with wildfly-swarm, and the way in which the "resource_access" claim would have to work would be along the lines of the draft you reference as you need use some portion of the service endpoint uri to represent the roles/group claim one is trying to grant. It is requiring quite a deep integration with the container security layers, so I'm inclined to drop the "resource_access" claim for 1.0 and revisit it as needed.

spec/src/main/asciidoc/refimpl.asciidoc
```diff
@@ -0,0 +1,37 @@
...
+
+* a default implementation of the JWTPrincipal interface
+* a JAX-RS ContainerRequestFilter prototype
+* JSR-375 IdentityStore and Credential prototypes
+* TODO: a JSR-196 ServerAuthModule
+
+### Wildfly Swarm Implementations
+
+A reference implementation for a custom auth-method of MP-JWT with the Wildfly/Wildfly-Swarm Undertow web container
+available under the ASLv2 license can be found at https://github.com/starksm64/microprofile-jwt-auth-wfswarm.
```
johnament: There is no "ASLv2" license

tck/pom.xml
```diff
@@ -50,59 +54,133 @@
     <dependencies>
         <dependency>
             <groupId>org.eclipse.microprofile.apis</groupId>
-            <artifactId>microprofile-config_1.0_api</artifactId>
+            <artifactId>microprofile-jwt-auth_1.0_api</artifactId>
```
johnament: artifactId should just be `microprofile-jwt-auth`

tck/pom.xml
```diff
@@ -50,59 +54,133 @@
...
-            <scope>provided</scope>
         </dependency>
 
         <dependency>
-            <!-- actually only referenced in JavaDoc -->
-            <groupId>org.apache.geronimo.specs</groupId>
-            <artifactId>geronimo-annotation_1.2_spec</artifactId>
-            <version>1.0</version>
+            <groupId>javax</groupId>
+            <artifactId>javaee-api</artifactId>
```
johnament: We can't depend on the full EE API.

spec/src/main/asciidoc/refimpl.asciidoc
```diff
@@ -0,0 +1,37 @@
+//
```
johnament: I don't think we should list out the reference impls.

johnament:
I'm going to merge this to unblock anything, but it still needs work.  I'll raise follow up issues.

PR #12, opened 2017-08-04
Further TCK updates

starksm64:


tck/src/test/java/org/eclipse/microprofile/jwt/tck/servlet/ServiceServlet.java
```diff
@@ -0,0 +1,101 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+import org.eclipse.microprofile.jwt.JWTPrincipal;
+import org.eclipse.microprofile.jwt.tck.ejb.IService;
+
+@HttpConstraint(rolesAllowed={"Tester"})
+@WebServlet("/ServiceServlet/*")
+public class ServiceServlet extends HttpServlet {
+    @EJB
+    private IService serviceEJB;
```
johnament: We can't rely on EJB.  Can't you use `@Inject` instead?

starksm64:
I have redone the TCK harness to come up with one approach to have the TCK run the required tests with a container profile that delegates to an arquillian based tck extension artifact a vendor must provide, for example:

mvn -Pcontainer -Dtck.container.groupId=org.wildfly.swarm -Dtck.container.artifactId=jwt-auth-tck -Dtck.container.version=1.0-SNAPSHOT test

johnament:
I'm merging as is, knowing that there's still work to do as far as dependencies go, e.g. no `web.xml`

PR #16, opened 2017-08-07
Some ideas for JWTPrincipal interface

hendrikebbers:
Some changes:
- removed public keyword since it is not needed in interface
- added default implementation for all the getter methods that returns a specific claim
- changed the return value of the method to get claim by name to a generic return value. By doing so you do not need to cast all the time
- changed primitive `long` return values for claims to `Long`. All the claims are not mandatory (as far as I know) and based on this you can end in a `NPE` based on auto boxing here if you use primitive `long`
- Changed `Set` return value to `Stream`. All the methods should return an unmodifiable collection and here `Set` is not the best interface since it provides methods to add entries. Sadly Java still has no interfaces for unmodifiable collections :( . But based on this maybe `Stream` is a good return value.
- Added `boolean containsClaim(String claimName)` convenience method. This is highly recommend if `Stream` is used instead of `Set` as return value for collections.
- Added a method that returns `Optional` for a claim.

All this are just some ideas of thinks that could be changed for the interface. Please let me know what you think about this :)

Additional questions: 
- Does it makes sense if the  JWTPrincipal interface extend Serializable ?
- KeyCloak provide some more specific claim types: [JsonWebToken.java](https://github.com/keycloak/keycloak/blob/d77824ff982f387828a04d5c04d7d1d82b4ff7a5/core/src/main/java/org/keycloak/representations/JsonWebToken.java) Does it makes sense to add them here?

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/jwt/JWTPrincipal.java
+++ b/api/src/main/java/org/eclipse/microprofile/jwt/JWTPrincipal.java
@@ -19,8 +19,10 @@
  */
 package org.eclipse.microprofile.jwt;
 
+import java.io.Serializable;
 import java.security.Principal;
-import java.util.Set;
+import java.util.Optional;
+import java.util.stream.Stream;
 
 /**
  * A read-only interface for the the claims required by Eclipse MicroProfile conforming tokens. Additional information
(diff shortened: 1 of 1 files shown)


hendrikebbers:
Oh, this PR is not made to merge it ;) It is only a collection of ideas.

starksm64:
These suggestions have been incorporated into #21 with modifications as discussed on the forum.

PR #17, opened 2017-08-08
Additional TCK cleanup/updates

starksm64:
- Begin adding release information to root pom
- Add the basic JWTPrincipal token parsing test and rename weld-se to tokens-se as CDI is not used. This drops weld-se dependency.
- Update the TCK run instructions to discuss both profiles

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,6 +43,47 @@
         </license>
     </licenses>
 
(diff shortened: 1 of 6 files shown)


starksm64:
This is now obsolete

PR #33, opened 2017-09-02
Allow claim to target types (for non-portable behavior).

### LINKED ISSUE #28: Remove target from Claim
For better reuse in many use cases (beyond whats defined in MP) it would be good if Claim did not specify a target, which would allow implementations more freedom on where users could put the claim.

### LINKED ISSUE #32: Should claims be non-binding
I suspect that the Claim qualifier should be nonbinding. https://docs.jboss.org/cdi/api/1.0-SP2/javax/enterprise/util/Nonbinding.html

Basically, the current impl requires registering beans of each claim type to satisfy injection points.  By making the attributes nonbinding, you can satisfy in a single producer method.

johnament:
Make claim attributes nonbinding to allow producers.

This closes #28
This closes #32

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/jwt/Claim.java
+++ b/api/src/main/java/org/eclipse/microprofile/jwt/Claim.java
@@ -19,6 +19,7 @@
  */
 package org.eclipse.microprofile.jwt;
 
+import javax.enterprise.util.Nonbinding;
 import javax.inject.Qualifier;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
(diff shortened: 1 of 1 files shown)


Emily-Jiang:
Merged this as this was what I am asking for as well.

johnament:
Even though it had the label `wip` on it?

johnament:
Now that this is merged, my tests all fail due to the attribute being non-binding.  For performance reasons, leaving the attribute as binding was working well.  I'll have to redo a lot of work to reset it.

Did you actually need it to be `@Nonbinding` @Emily-Jiang ?  Or did you need the target change?

Emily-Jiang:
I need it to be nonbinding so that the app can start ok. Do you provide a list of beans with the predefined matching qualifiers? Why did your test fail? Feel free to revert your previous changes and discuss again.

johnament:
Hammock was registering a bean per `Claim` qualifier instance.  With this change, I just need to register a single bean.  But it means I have a performance hit of handling a per type mapping within a single bean.

Emily-Jiang:
I am ok for you to revert the change. Since there is a defined list of values, it should not be too difficult to specify more beans. Thoughts?

johnament:
At this point, its fine.  I've changed to align.  I'm assuming you have this working in wildfly as well @starksm64 ?

PR #42, opened 2017-09-13
Make 'aud' optional

dblevins:
We aren’t providing any guidance for its use, so people will have to “fill in something” because we have it as a requirement for interoperability and we are technically required to reject JWTs without it.  The concern is since we are forcing people to put something there, unsupervised, we will end up with something like Java EE's “mappedName” which is effectively a vendor-specific bucket we’re all required to support — a bucket we can’t further specify later because it might break what people are using it for which was never specified.

As the JWT spec also has it as optional, it feels like it adds potential interoperability cost which could limit us.  I’ve moved it into the custom claims section of the spec and TCK like we did for roles.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
+++ b/api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
@@ -60,15 +60,6 @@ public interface JsonWebToken extends Principal {
     }
 
     /**
-     * The aud(Audience) claim identifies the recipients that the JWT is
-     * intended for.
-     * @return the aud claim.
-     */
-    default Set<String> getAudience() {
-        return getClaim(Claims.aud.name());
-    }
-
-    /**
      * The sub(Subject) claim identifies the principal that is the subject of
      * the JWT. This is the token issuing
      * IDP subject, not the
--- a/api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
+++ b/api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
@@ -60,6 +60,15 @@ public interface JsonWebToken extends Principal {
     }
 
     /**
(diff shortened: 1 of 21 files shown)


api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
```diff
@@ -60,15 +60,6 @@ default String getIssuer() {
     }
 
     /**
-     * The aud(Audience) claim identifies the recipients that the JWT is
-     * intended for.
-     * @return the aud claim.
-     */
-    default Set<String> getAudience() {
```
johnament: Why remove the default method?

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/ClaimValueInjectionEndpoint.java
```diff
@@ -146,30 +143,7 @@ else if(jtiValue.equals(jwtID)) {
             .build();
         return result;
     }
-    @GET
-    @Path("/verifyInjectedAudience")
-    @Produces(MediaType.APPLICATION_JSON)
-    public JsonObject verifyInjectedAudience(@QueryParam("aud") String audience) {
```
johnament: Maybe its me, or my interpretation of what `aud` was for, but it seems like all we're checking is that you support aud.  So what is the issue with just supporting aud (but not verifying it)?

dblevins: Section 4.1 definitely says (said) the JWT itself must have the `aud` claim to be considered an interoperable MP JWT.

In terms of why use the `getClaim(String)` method, that's the pattern we've taken for all other non-required claims so I followed it for consistency.

PR #96, opened 2018-06-13
cleanup and comments

jeanouii:
Hey Scott, previous PR did not pass the IP check because the email was not the one I used for the ECA. I tried to set the good email for commits on this project and looks like there is no error regarding the IP check. Would you like to check if that's good or not?


sberyozkin:
@jeanouii Hi, can you please update this PR ? Thanks

PR #99, opened 2018-07-24
Fixing scope and typo

ivanjunckes:
It looks like an issue in the spec where it says, the implementation should only support @RequestedScoped but then it shows @ApplicationScoped in the code.

Signed-off-by: ivanjunckes <ijunckes@tomitribe.com>


ivanjunckes:
I now understand why is ApplicationScoped there, so closing the issue.

PR #102, opened 2018-08-17
JWT Auth TCK improvement

jGauravGupta:
Signed-off-by: Gaurav Gupta <gaurav.gupta.jc@gmail.com>


starksm64:
Ok, looks good except that we want to retain the test that a java.security.Principal can be injected. As noted on #104 it would be good to have an additional test that validates the SecurityContext principal type, so I'll merge this and add back in the java.security.Principal injection test.

Issue #109, opened 2018-08-31
MP JWT to produce Json Web Token


Emily-Jiang:
At the moment, different vendors choose different ways to produce Json Web Tokens,. It does not help with app's portability (contradicts with MP's portability mission). Should we define a common way to generate Json Web Token and maybe make MP as a token issuer?

Emily-Jiang:
Talked this on today's hangout. MP JWT can create a factory to abstract the Token creation layer and then microservice developer can just call the factory to get a JSON web token. This is a great abstract layer to hide the internal differences among different vendors to make microservices portable.

sberyozkin:
Hi, but tokens are issued by the authorization servers ?

hrstoyanov:
I do need to create token out of an httt request, see https://github.com/OpenLiberty/open-liberty/issues/6087

sberyozkin:
@hrstoyanov but it is not about issuing the tokens. The token is either available in the HTTP request or not, if yes - it was either issued by the OIDC provider or self-issued by the client. I'd say your issue is more related to  #120.
For ex, issuing the tokens (access, id or refresh tokens), managing them as part of various SSO flows, etc, is all a complex process, which is not something  MP JWT should get into IMHO

sberyozkin:
Though having a builder style API for generating the self-issued tokens as I believe implied by @Emily-Jiang may be a nice enhancement. I might give it a go and do a PR a bit later on

sberyozkin:
A bit of a challenge there is that JsonWebToken is, in case of MP JWT, an instance of JWS, and may also become an instance of JWE. That said this would be API should be as simple as possible, as opposed to becoming a complete JOSE builder API...have claim setters plus either the signing or encrypting (keys)

hrstoyanov:
@sberyozkin 
The question is why can't we do:
```
@Inject
Optional<JsonWebToken> token;
```

because the presence of JWT is optional in the request headers.

We can inject Optional config properties with MP-CONFIG, so why not the same for JWT?

sberyozkin:
@hrstoyanov I'm not disagreeing with you :-), just saying IMHO this is not relevant to this specific issue

hrstoyanov:
@sberyozkin 
Sorry then. I do remember filing #120 , not sure when it will be included int he MP JWT specs

sberyozkin:
Hi All
I've started working on the API which I'm hoping to make available for the discussion and the team review in January 2020. Thanks.

[...]

PR #122, opened 2019-02-06
add encryption utility

brutif:
This probably doesn't quite fit with conventions but hopefully is sufficient to start.

### DIFF
--- a/sandbox/pom.xml
+++ b/sandbox/pom.xml
@@ -13,15 +13,18 @@
     specific language governing permissions and limitations
     under the License.
 -->
+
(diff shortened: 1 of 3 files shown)


sberyozkin:
I wonder if the builder API for producing either signed or encrypted or signed and then encrypted JWTs is really the way to go, see #109. The implementations will have a choice which JOSE library to use, etc

sberyozkin:
Hi @brutif Thank you for this PR but I'm going to close it as we'll try to come up with the standard builder API for securing the tokens shortly. Cheers !

Issue #118, opened 2018-11-16
TCK Challenge: Tests using URL location make a too restrictive assumption


tomas-langer:
Tests
1. PublicKeyAsPEMLocationURLTest
2. PublicKeyAsJWKLocationURLTest
Make an assumption that the application is deployed at the time the JWT handling is configured.
If an implementation configures the JWT Auth handling at startup (as we do) and the key is not available at that time, the startup fails and the application never gets deployed.
The specification does not define WHEN the URL should be reached, whether it can/should be cached etc., so the implementation may choose. 
As we follow the "fail fast" principle, we fail if the configuration cannot be completed, thus never deploying the application. As a result these TCK tests cannot be executed, even though the implementation is valid and according to the specification.

radcortez:
+1

radcortez:
My understanding on the spec is that the first load / key validation needs to happen at deploy time. From the spec:

_MicroProfile JWT implementations are required to throw a `DeploymentException` if the JWK
`kty` field is missing or JSON text is found, but does not follow either JWK or JWKS
format._

It doesn't make sense to throw a DeploymentException if you load the key after the application is deployed.

Anyway, on TomEE we have the same issue. If the key is part of the deployment, we are unable to reach an endpoint before the application is completely deployed. After that point, as I said, it doesn't make sense to throw a DeploymentException.  

Also, I believe on a proper usage scenario, you will not be exposing the JWK Key on the same application that needs it. So, on the TCK, I think we can easily fix this by adding another Arquillian deployment to expose the key and have the main deployment call it to run the test. Here is how we have done it:
https://github.com/apache/tomee/pull/173/commits/4dee78b989bd7c85226938d9dce0971e5a1e8a1f#diff-7855d23518772909eb5be87d68093b0b

I'm happy to submit a PR to the TCK to achieve the same result.

starksm64:
Yes, agreed.

@radcortez Sure, create a PR to fix the test.

sberyozkin:
@radcortez Hey Roberto, so please do this PR :-)

radcortez:
Ops, sorry. I guess I've forgotten about this. I'll put this on my priority list :)

radcortez:
Ok, here is a fix for it:
https://github.com/eclipse/microprofile-jwt-auth/pull/148

On the other hand, I've started to think about this and I'm not completely sure if this should be the right behaviour from a spec point of view.

[...]

PR #148, opened 2020-01-31
Fix for #118. Provide separate deployment to retrieve a key from a URL.

radcortez:


sberyozkin:
@radcortez Hi Roberto thanks for this PR, sorry for a delay. I'm assuming this update will basically make sure that the test will run irrespective of how the application is deployed and its keys are initialized, correct ?

radcortez:
Yeah, the idea over here is that by using Arquillian multiple deployments, you could deploy an initial application that exposes the key and the second application that consumes the key depends on the deployment of the key app.

sberyozkin:
Hi @radcortez OK thanks. I've realized that some application servers (supporting single deployments like Thorntail or any other container which can't handle multiple deployments) may have problems. Perhaps we may split this test or come up with some other idea, otherwise I'll just mark it for `JWT 2.0` where only `UnresolvableKeyException` without a tie-in to the time the keys are initialized, is checked...

radcortez:
Sure.

Another option could be to start a separate http server via `HttpServer` and expose the key over there. Not sure if we can plugin some sort of awaitability to wait for the server to get up and then deploy the Arquillian stuff. Might work.

I'll give it a try.

sberyozkin:
@radcortez It is a nice and neat idea, really is IMHO. I have a reservation now though about the issue validity. If the server fails at the startup if no remote location is available then it means it is not able to get hold of the rotated keys which is only possible by periodically refreshing a JWK set. If we have this HTTP server then what exactly we will test ? I mean it won't resolve the ambiguity about throwing or not throwing DeploymentException if no remote location is available.
Perhaps for 2.0 it could make sense to have such a test with the HTTP server emulating the case of returning a set with a non-matching kid, returning 404, etc, to support few variations around the remote key unavailability and verify that no matter what the client gets the same HTTP error, etc

radcortez:
Well, right now this was just to fix the issue for certain servers where the endpoint you use for the key is not available to query, because is part of the current deployment (and egg and chicken problem).

I agree that we need to discuss this further and figure out what should be the correct behaviour of the spec in these cases (404, wrong kid, etc).

Issue #58, opened 2018-01-20
Define minimum useful support for JWE


starksm64:
Would like to see if we can have interoperable encrypted JWT, maybe some minimum support requirement similar to OIDC, http://openid.net/specs/openid-connect-core-1_0.html#Encryption

JWE: https://tools.ietf.org/html/draft-ietf-jose-json-web-encryption-40

starksm64:
JWK and headers need to be looked at.

sberyozkin:
You are right, but it will mean the private keys manages by the AS will have to be exported into all of RS. Which is probably a key management challenge, though yes, out of scope for MP-JWT :-)

sberyozkin:
I'm wiring the encryption support into the builder API I'm working upon

chunlongliang-ibm:
In addition to JWT API, we should consider to mpJWT configuration to decrypt JWE  for some common and simple scenarios, in particular, the dir ( symmetric key) scenario. Only JWE with 'dir' is supported.
1. The 'dir' alg is much simpler to be processed, and the symmetric key could be defined via mp-config.
2. The 'dir' scenario is also useful given that we now plan to support JWT builder API. If client builds self-signed JWT for downstream service, it is reasonable to assume both should share some secret, which can be used as encryption key.

PR #149, opened 2020-02-12
Introduce JWT Builder API

sberyozkin:


api/src/main/java/org/eclipse/microprofile/jwt/algorithm/SignatureAlgorithm.java
```diff
@@ -0,0 +1,45 @@
...
+package org.eclipse.microprofile.jwt.algorithm;
+
+/**
+ * JWT JSON Web Signature Algorithms.
+ * 
+ * @see <a href="https://tools.ietf.org/html/rfc7518#section-3">https://tools.ietf.org/html/rfc7518#section-3</a>
+ */
+public enum SignatureAlgorithm {
+    RS256,
+    RS384,
```
rdebusscher: What is the purpose of the different signing algorithms when in the JWT Spec itself only RS256 is mentioned?

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/KeyEncryptionAlgorithm.java
```diff
@@ -0,0 +1,47 @@
...
+    private KeyEncryptionAlgorithm(String algorithmName) {
+        this.algorithmName = algorithmName;
+    }
+
+    public String getAlgorithm() {
+        return algorithmName;
+    }
+
+    public static KeyEncryptionAlgorithm fromAlgorithm(String algorithmName) {
+        return KeyEncryptionAlgorithm.valueOf(algorithmName.replaceAll("-", "_").replaceAll("\\+", "_"));
```
rdebusscher: This logic fails when passing a parameter which doesn't correspond to valid valid or when null is passed.

This is much more safer

return Arrays.stream(KeyEncryptionAlgorithm.values()).filter(e -> e.algorithmName.equals(algorithmName))
                .findAny().orElse(null);

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/KeyEncryptionAlgorithm.java
```diff
@@ -0,0 +1,47 @@
...
+public enum KeyEncryptionAlgorithm {
+    RSA_OAEP("RSA-OAEP"),
+    RSA_OAEP_256("RSA-OAEP-256"),
+    ECDH_ES("ECDH-ES"),
+    ECDH_ES_A256KW("ECDH-ES+A256KW"),
+    A256KW("A256KW");
+
+    private String algorithmName;
+
+    private KeyEncryptionAlgorithm(String algorithmName) {
```
rdebusscher: private keyword can be removed as constructor is always private for enum.

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/SignatureAlgorithm.java
```diff
@@ -0,0 +1,45 @@
...
+    HS256,
+    HS384,
+    HS512;
+
+    public String getAlgorithm() {
+        return this.name();
+    }
+
+    public static SignatureAlgorithm fromAlgorithm(String algorithmName) {
+        return SignatureAlgorithm.valueOf(algorithmName);
```
rdebusscher: see KeyEncryptionAlgorithm remark and solution.

api/src/main/java/org/eclipse/microprofile/jwt/builder/Jwt.java
```diff
@@ -0,0 +1,93 @@
...
+ * String = Jwt.claims("/tokenClaims.json").sign();
+ * </code>
+ * </pre>
+ * <p>
+ * The next example shows how to use {@link JwtClaimsBuilder} to add the claims and encrypt a JSON
+ * representation of these claims with a configured encrypting key:
+ * 
+ * <pre>
+ * <code>
+ * String = Jwt.claims().issuer("https://issuer.org").claim("custom-claim", "custom-value").encrypt();
```
rdebusscher: encryption is not supported by MP JWT spec yet, so it this already useful?

api/src/main/java/org/eclipse/microprofile/jwt/builder/Jwt.java
```diff
@@ -0,0 +1,93 @@
...
+        return JwtProvider.provider().claims(claims);
+    }
+
+    /**
+     * Creates a new instance of {@link JwtClaimsBuilder} from a JSON resource.
+     * 
+     * @param jsonLocation JSON resource location
+     * @return {@link JwtClaimsBuilder}
+     */
+    public static JwtClaimsBuilder claims(String jsonLocation) {
```
rdebusscher: Support for URI so that resource within archive can be used? (within a jar contained within the WAR, for example)

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtClaimsBuilder.java
```diff
@@ -0,0 +1,153 @@
...
+ */
+package org.eclipse.microprofile.jwt.builder;
+
+import java.util.Set;
+
+/**
+ * JWT Claims Builder.
+ * 
+ * <p>
+ * JwtClaimsBuilder implementations must set the 'iat' (issued at time), 'exp' (expiration time)
```
rdebusscher: So when we set the iat in the call to Jwt.claims(Map<>) call, it needs to use that value and ignore the one the developer sets through JwtClaimsBuilder. issuedAt().

I would revert that logic or even disallow setting 'registered' claims through the Jwt.claims(Map<>) call.

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtClaimsBuilder.java
```diff
@@ -0,0 +1,153 @@
...
+     */
+    JwtClaimsBuilder preferredUserName(String preferredUserName);
+
+    /**
+     * Set an issuedAt 'iat' claim
+     * 
+     * @param issuedAt the issuedAt time in seconds
+     * @return JwtClaimsBuilder
+     */
+    JwtClaimsBuilder issuedAt(long issuedAt);
```
rdebusscher: Is this the actual value of the iat or the 'timeStamp'?

Since iat must be in seconds so roughly `new Date().getTime()/1000`

Add overloaded method for Date, LocalDate, ... ?

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtClaimsBuilder.java
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * Set an expiry 'exp' claim
+     * 
+     * @param expiredAt the expiry time
+     * @return JwtClaimsBuilder
+     */
+    JwtClaimsBuilder expiresAt(long expiredAt);
```
rdebusscher: same remark as with issuedAt.

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtClaimsBuilder.java
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Set a custom claim. Claim value is converted to String unless it is
+     * an instance of {@code Boolean}, {@code Number}, {@code Collection}, {@code Map},
+     * {@code JsonObject} or {@code JsonArray}.
+     * 
+     * @param name the claim name
+     * @param value the claim value
+     * @return JwtClaimsBuilder
+     */
+    JwtClaimsBuilder claim(String name, Object value);
```
rdebusscher: Does this method also allow to set `iat`, `exp` etc ... ?

And if so, is there a correct conversion from Dates for example?

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtClaimsBuilder.java
```diff
@@ -0,0 +1,153 @@
...
+     * @return JwtSignatureBuilder
+     */
+    JwtSignatureBuilder jws();
+
+    /**
+     * Set JsonWebEncryption headers and encrypt the claims by moving to {@link JwtEncryptionBuilder}
+     * 
+     * @return JwtSignatureBuilder
+     */
+    JwtEncryptionBuilder jwe();
```
rdebusscher: There is no support for JWE in spec at this moment.

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtEncryption.java
```diff
@@ -0,0 +1,80 @@
...
+     * which can point to a PEM, JWK or JWK set keys.
+     * 'RSA-OAEP-256' key and 'A256GCM' content encryption algorithms will be used
+     * unless different ones have been set with {@code JwtEncryptionBuilder}.
+     * A key of size 2048 bits or larger MUST be used with the 'RSA-OAEP' and 'RSA-OAEP-256' algorithms.
+     *
+     * @param keyLocation the location of the keyEncryptionKey which encrypts the content encryption key
+     * @return encrypted JWT token
+     * @throws JwtEncryptionException the exception if the encryption operation has failed
+     */
+    String encrypt(String keyLocation) throws JwtEncryptionException;
```
rdebusscher: Support for URI?

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/ContentEncryptionAlgorithm.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
```
rdebusscher: Years are incorrect

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtEncryption.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.crypto.SecretKey;
+
+/**
+ * JWT JsonWebEncryption.
+ */
+public interface JwtEncryption {
+
+    /**
+     * Encrypt the claims or inner JWT with {@link PublicKey}.
+     * 'RSA-OAEP-256' key and 'A256GCM' content encryption algorithms will be used
```
rdebusscher: RSA-OAEP-256 is only for using the RSA Public Key, A256GCM only for an AES key (SecretKey)

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtEncryption.java
```diff
@@ -0,0 +1,80 @@
...
+     * unless different ones have been set with {@code JwtEncryptionBuilder}.
+     * A key of size 2048 bits or larger MUST be used with the 'RSA-OAEP' and 'RSA-OAEP-256' algorithms.
+     *
+     * @param keyEncryptionKey the key which encrypts the content encryption key
+     * @return encrypted JWT token
+     * @throws JwtEncryptionException the exception if the encryption operation has failed
+     */
+    String encrypt(PublicKey keyEncryptionKey) throws JwtEncryptionException;
```
rdebusscher: PublicKey also allows ECPublicKey

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtSignatureBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+ * <p>
+ * Note that JwtSignatureBuilder implementations are not expected to be thread-safe.
+ * 
+ * @see <a href="https://tools.ietf.org/html/rfc7515">RFC7515</a>
+ */
+public interface JwtSignatureBuilder extends JwtSignature {
+
+    /**
+     * Set a signature algorithm.
+     * Note that only 'RS256' (default), 'ES256' and 'HS256' algorithms must be supported.
```
rdebusscher: Again much broader then current Spec?

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtSignatureBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+     * Sign the claims with {@link SecretKey} and encrypt the inner JWT by moving to {@link JwtEncryptionBuilder}.
+     * 
+     * @param signingKey the signing key
+     * @return JwtEncryptionBuilder
+     * @throws JwtSignatureException the exception if the inner JWT signing operation has failed
+     */
+    JwtEncryptionBuilder innerSign(SecretKey signingKey) throws JwtSignatureException;
+
+    /**
+     * Sign the claims with a key loaded from the location set with the "smallrye.jwt.sign.key-location" property
```
rdebusscher: smallrye mentioned here.

api/src/main/java/org/eclipse/microprofile/jwt/builder/spi/JwtProvider.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Service provider for JWT Claims Builder objects.
+ *
+ * @see ServiceLoader
+ */
+public abstract class JwtProvider {
+    /**
+     * Name of the default {@code JwtProvider} implementation class.
+     */
+    private static final String DEFAULT_JWT_PROVIDER = "io.smallrye.jwt.build.impl.JwtProviderImpl";
```
rdebusscher: smallrye

sberyozkin:
@rdebusscher Hi, thanks for the review so far, I'll deal with you individual comments later, here is feedback on your first general one

> I did only a brief review of the PR, I did not check the API in detail (but I assume this is already in SmallRye and thus more or less proven to be useable)
We have Quarkus users trying it yes.

> Important question is why we define an API to generate any kind of JWT (JWS and JWE with many options) when there are already at least 3 libraries which does the same.

There is some mix up here. It does not matter how many libraries are around, the goal is to enrich MP JWT with its own `portable` JWT generation API as proposed by @Emily-Jiang at #109 (as well as by my team colleagues) as opposed to everyone using their own snippets.

Note there is also a [task](https://github.com/eclipse/microprofile-jwt-auth/issues/58) to support MP JWT accepting the JWE-protected tokens. Hence the JWE option. 

> One remark from the past was that developers should not generate their own JWT tokens, this is something the authentication provider (like Keycloak) does. 

I myself have been issuing such remarks but MP JWT has not been designed as an OIDC adapter, i.e the library that must be accepting the tokens as part of OIDC flows only and the reality is right now the developers use it for all sort of JWT token issuance scenarios, including self-issuing, with every developer code using, well, some specific library, which is what this API is about to address.

> And if you really need such an API, I see only room for a very minimal API which generates JWTs indicated in the spec.

I disagree with it. It has not played well for MP JWT to require that `groups` for example must be there or `upn` as many certified OIDC servers don't set them, while it is in the interest of MP JWT to accept the tokens from as many providers as possible.

Likewise MP JWT Generation API should see further than just issuing MP-JWT tokens with the limited support which are only meant for MP JWT servers, this is a close world assumption. 
This API is really very simple and will help with making MP JWT more visible to a larger develoiper audience. It is very simple really, `sign`, `encrypt` and `inner-sign/encrypt`, the latter two options will be part of #58 on the JWE verification side as both options are used by Jose libraries.

Also missing an extensive set of TCKs to test the implementations of this API.

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/SignatureAlgorithm.java
sberyozkin: @rdebusscher Because the JWT world is not only about MP JWT servers (same as the JWT tokens are not always issued with `upn` and `groups`) and it is in the interests of MP JWT IMHO for it to be usable by a larger developer audience. 
For example, even an ultra secure MP JWT server may want `RS512`. And we have a practical issue with one of Amazon services repackaging an `RS256` signed token with `ES256` as reported by our user.

api/src/main/java/org/eclipse/microprofile/jwt/builder/Jwt.java
sberyozkin: It will be supported, see #58. So we need to expect it, and once again, MP JWT should be able to reach beyond MP JWT servers, same as we want MP JWT servers to accept as many tokens as possible

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtEncryption.java
sberyozkin: That is what that comment implies. the former is for the encrypting the content key, the latter is for this content key to encrypt the actual content

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtEncryption.java
sberyozkin: Right, this allows for Elliptic Curve signatures. See my above comment too, we have a practical case wiuth ES256 signed MP JWT tokens, and yet again, it is only about targeting MP JWT servers :-)

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtSignatureBuilder.java
sberyozkin: See above.

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtSignatureBuilder.java
sberyozkin: @rdebusscher Is it not acceptable ? A default implementation is typically referenced in the ServiceLoader loaded implementations, JSONB for example, etc.
@kenfinnigan @Emily-Jiang what is the guideline here ?

api/src/main/java/org/eclipse/microprofile/jwt/builder/spi/JwtProvider.java
sberyozkin: Will follow the resolution as per the above comment

sberyozkin:
@rdebusscher Replied to all the comments. Thanks for the comprehensive feedback, lets keep going :-)

rdebusscher:
I know there is an open task for JWE, so that should be decided and implemented in the spec firest before we should be think of creating a JWE.

An open spec doesn't work. Because if there is tomorrow a new signing algorithm, it is assumed to be supported automatically by this API which is not the cause of course.
Within a spec there needs to be a very detailed description of what algorithms, encryption methods and type of keys are valid.

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/SignatureAlgorithm.java
rdebusscher: Assume the open spec as you propose, then the list of algorithms is missing

PS256, PS384, PS512, ES256K, and EdDSA (and possibly missing others)

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/ContentEncryptionAlgorithm.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * * JWT JSON Web Content Encryption Algorithms.
+ * 
+ * @see <a href="https://tools.ietf.org/html/rfc7518#section-5">https://tools.ietf.org/html/rfc7518#section-5</a>
+ */
+public enum ContentEncryptionAlgorithm {
+    A256GCM("A256GCM"),
+    A128CBC_HS256("A128CBC-HS256"),
+    A256CBC_HS512("A256CBC-HS512");
```
rdebusscher: Also here the list is missing encryption Methods if we go for an open spec. (and enum doesn't allow easy extension for new encryptions)

A128GCM, A192GCM, A192CBC-HS384

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/KeyEncryptionAlgorithm.java
```diff
@@ -0,0 +1,47 @@
...
+ * JWT JSON Web Key Encryption (Management) Algorithms.
+ * 
+ * @see <a href="https://tools.ietf.org/html/rfc7518#section-4">https://tools.ietf.org/html/rfc7518#section-4</a>
+ */
+public enum KeyEncryptionAlgorithm {
+    RSA_OAEP("RSA-OAEP"),
+    RSA_OAEP_256("RSA-OAEP-256"),
+    ECDH_ES("ECDH-ES"),
+    ECDH_ES_A256KW("ECDH-ES+A256KW"),
+    A256KW("A256KW");
```
rdebusscher: And missing (in the open world scenario)

A128KW, A192KW, dir, ECDH-ES+A128KW, ECDH-ES+A192KW, A128GCMKW, A192GCMKW, A256GCMKW, PBES2-HS256+A128KW, PBES2-HS384+A192K, PBES2-HS512+A256KW. (and possibly others)

sberyozkin:
> I know there is an open task for JWE, so that should be decided and implemented in the spec firest before we should be think of creating a JWE.

I don't think that it should be decided and implemented first.

> An open spec doesn't work. Because if there is tomorrow a new signing algorithm, it is assumed to be supported automatically by this API which is not the cause of course.

You haven't looked carefully enough. The algorithms are typed and only those which are well known in the JWA spec are typed. Furthermore, only a bare minimum if required to be supported. The rest - is optional.

> Within a spec there needs to be a very detailed description of what algorithms, encryption methods and type of keys are valid.

Exactly, this will be done for the minimum required set

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/KeyEncryptionAlgorithm.java
sberyozkin: @rdebusscher I'm sorry, this comment is not productive. There is an effort there to limit a set of algorithms to only those which are `recommended` by the JWA spec. New ones can be added upon the demand if needed

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/ContentEncryptionAlgorithm.java
sberyozkin: See the comment below. This API is open enough, but has been carefully looked at. Lets focus on the quality of the API. Please take the latest smallrye library just to have a feel for it, IMHO it is quite good and will be good for MP JWT

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/KeyEncryptionAlgorithm.java
sberyozkin: @rdebusscher I'll be open to trimming this list but these algorithm enums will not put any burden on the implementations 100%. The goal is type the minimum list of `required` algorithms from the JWA spec, so that users don't type the algorithm names manually, and, at the MP JWT level, we'll only require to support the very few ones.
Signature: `RS256` (easy:-) ), `ES256` - you won't even have to change your impl code to support it and it is added to support MP JWT servers accepting such tokens. `HS256` - the users are signing the JWT cookies but also would be good to verify them with MP JWT. It is about extending MP JWT reach.

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/ContentEncryptionAlgorithm.java
rdebusscher: I find it an arbitrary decision which algorithms and encryption methods are supported on top of the one which can be used within MP JWT spec.

So in other words, if we only support those defined in the spec (like RS256 is the only mentioned) or all of them, that would make sense.

But now we have some of them and other not. (and not all of the JWA specified ones are now in the PR)

rdebusscher: And saying they can optionally is a bad thing. Implementation A will support all possible, while implementation B only the limited list.

This means an app which works on Implementation A fails on Implementation B which is an undesirable situation.

sberyozkin: @rdebusscher same is true for the line that other than 2048/1024 bits are supported optionally for the verification keys, right ?
That what I was implying. Saying a given algorithm is optional is a clear warning to the users that the interoperability is not guaranteed. But if a user has a requirement to create a token with something other than `RS256` using this API then they will have an option to use this API, assuming the given implementation is ready to do it. If this API does not have a typed support for a given algorithm then the users will need to open an issue and justify one more enum value added. I've already justified to you why `ES256` and explained why `HS256` is added (we have the users manually generating the cookies and this where the API should help) 

Anyway, I'll trim few values for now, `RS384`/`RS512`, etc, but let me say again, I don't agree that MP JWT server requirements are the ones that must drive what may be in the generation API, there is no such requirement and it is wrong IMHO to assert it. It is not a good strategy to keep a closed world assumption exactly for the same reason as assuming that soon all the world will start issuing MP JWT compliant tokens and they don't then it is not the problem of MP JWT.

sberyozkin: > So in other words, if we only support those defined in the spec (like RS256 is the only mentioned) or all of them, that would make sense.
I've several time already, no, IMHO it would not make sense to support all of them, and it was not the intention. I went through the list and chose few required/recommended ones (as well based on some evidence that those algorithms may be used in practice) and this enums must not include optional ones, and if some of them made it there then I'll trim them.

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/KeyEncryptionAlgorithm.java
sberyozkin: @rdebusscher I've looked more carefully at this comment and I don't think it is a good idea, sorry about it. For few reasons: several of my colleagues have noted that the streaming API is expensive and it simply looks as if it is too much for the propose of converting an algorithm like `RSA-OAEP` into a valid enum value like `RSA_OAEP` which is the reason behind this code which is your code wold not do.
I'll make sure that NPE is avoided. And I prefer the exception thrown if no valid enum values exists

rdebusscher: ok, But String manipulations like replaceAll are also expensive. But fine for me.

sberyozkin:
@rdebusscher I've trimmed down the list of algorithms significantly; let me resolve some of the converstaions (feel free please to reopen as needed) and I will follow with some comments on the update. Thanks

sberyozkin:
Hi @rdebusscher,
- I've trimmed down the list of algorithms significantly, but relaxed the requirement that only those in these enums can be accepted; instead a warning is added that neither the portability nor interoperability can be guaranteed with the guidance that supporting any other algorithm is optional. IMHO it is a good compromise as we only offer a typed support for the algorithms which must be supported while we need to keep a window open for the users to request that other algorithms have a typed support as needed. I've simplified the enums too as a result.
- I've left only 2 values in `SignatureAlgorithm`, with `ES256` being the only addition (I can link to our project's issue to confirm there is a practical issue which can affect all MP JWT implementations which need to verify the tokens). I'm going to follow with a small PR syncing MP JWT server requirements to enable ES256 on the server too. It will be useful anyway IMHO to start slowly extending the algorithm coverage. I've dropped `HS256`/etc as syncing up on the server is problematic at the moment so we'll need a separate dedicate effort for the symmetric keys.
- encryption - only `RSA-OAEP` (since it has a `recommended+` JWA status) key management and `AGCM256` content encryption algorithms are required. I can sync easily enough with a follow up PR for the server requirements, it would be really a single new property, with the verify key becoming optional on the JWE path unless the nested token is available.

Thanks

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/ContentEncryptionAlgorithm.java
sberyozkin: What should we put there, `2020-2021` ?

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtClaimsBuilder.java
sberyozkin: @rdebusscher Sorry, don't think so. There is no ambiguity. If the builder is set with some claim then unless it has been reset with a `claim` or typed claim setter it remains as is, This doc mentions that that if no `iat` is available then it must be provided OOB for the users to avoid doing the boilerplate code when the defaults suit.

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtEncryption.java
sberyozkin: The rules are the same as for `mp.jwt.verify.public-key-location`, not sure we need to overload here

api/src/main/java/org/eclipse/microprofile/jwt/builder/Jwt.java
sberyozkin: Or should we just update the docs to note it can be a location pointing to a class path, filesystem or the HTTP location. If we introduce `URI` then `String` would need to be restricted as to what it may contain and I'd rather follow the the existing MP JWT guideline (for the key location, etc) ?

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtClaimsBuilder.java
sberyozkin: The same what is done [here](https://github.com/eclipse/microprofile-jwt-auth/blob/master/tck/src/main/java/org/eclipse/microprofile/jwt/tck/util/TokenUtils.java#L134) is assumed in this case.

api/src/main/java/org/eclipse/microprofile/jwt/builder/Jwt.java
rdebusscher: Yes, I wanted to change parameter type from String to URI.

With String, anything can be put into it but we interpret it as a File path location.

rdebusscher:
Thanks @sberyozkin,

The restricted algo list and encryption makes sense and we don't try to support all the algos and methods which exist.

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/ContentEncryptionAlgorithm.java
rdebusscher: For new files, it should be just 2020. (no fork) 

For updated files, the 'end' year should be updated.

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtClaimsBuilder.java
rdebusscher: Ok, that was not clear for me from the Javadoc. (and my comment wasn't clear either I think, sorry)

So when a value is set by the initial creation with the map `Jwt.claims(Map<>)`, it will and can be overridden by a call to `JwtClaimsBuilder. issuedAt().`

If that is the expected behavior, I prefer this also.

api/src/main/java/org/eclipse/microprofile/jwt/builder/JwtClaimsBuilder.java
rdebusscher: But for me, the long parameter is not clear and unambiguous. You can use something like `now()` but also `new Date().getTime()/1000`.  But `iat` should be in seconds and that is not clear from the method signature.

So I propose to use `Date` and `LocalDate` as that is very clear and removes the knowledge requirement from the developer what the exact format of `iat` and `exp` are.

sberyozkin:
@rdebusscher Hi, may be of interest, right now I'll use smallrye-jwt to build a JWT to support a [client_secret_jwt](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) OIDC client authentication using `HS256`. This is one of the good cases where I'd prefer to use MP JWT in time :-), but for now I'll stick to the commitment not to push `HS256` as a required algorithm until a server-side sync is available. cheers

sberyozkin:
To-do list for myself:
- address the licence year issue, and most likely drop a `smallrye.jwt` reference
- properties should follow a pattern like `key.localtion` as opposed to `key-location` given that the former pattern is already in use in MP-JWT
- drop the requirement that an `innerSign()` call produces a `none` signature if `mp.jwt.encrypt.key.location` is missing because it hide a bug where the property was not configured by mistake and `none` can not meet the sign-then-encrypt expectations (non-repudiation, etc)

sberyozkin:
Marking for 2.0 for now. I'll see if I can find the time to complete it earlier, but it is pretty tight so 2.0 may be more appropriate, we'll see

dblevins:
Catching up on this thread.  I agree with @rdebusscher earlier comments that JWT creation has explicitly been identified in the past as out of scope.  Here were some of the reasons:

- There are plenty of client libraries that do this
- It reads either as an endorsement that MicroProfile encourages people to write their own authorization servers or that the MicroProfile implementation should be an authorization server
- Doing this properly is actually very very hard due to constant change, you need to be intimately involved in the JOSE specs and latest developments.

If one were created we'd want it to in some way be optional so that we'd be free to not go that direction.

sberyozkin:
@dblevins David, thanks, but you may have misunderstood the goal of the PR.

Instead of every developer having to hack some non-portable and non-controlled code this PR will give them a chance to write a code which works across multiple libraries, and this code will be TCK verified.

This API has been carefully designed such it does not require an academic qualification to impplement. I have a 1st hand experience in implementing JOSE so I know what I'm proposing, please don't take it as some over-confident remark. And please, don't say it is hard to implement :-), certainly not for the Tomeetribe team :-). I can share a link to the branch where I coded it on top of Jose4j in half a day.  

I'd like you to trust this API has not been baked in a rush - we have Quarkus users experimenting with it in our project and everyone I've talked to about it is happy,

The other important point which has been discussed earlier is that it is pretty important to open MP JWT both ways for it to appeal to a larger audience - for it not only be a server but become a producer and also show some lead in supporting not only signed but signed and/or encrypted tokens.

CC @rdebusscher

sberyozkin:
@dblevins I'll have no problems making it optional for the implementers to support, so if TomeeTribe would prefer recommending its uses to code directly against Jose4J or other lower level library then I'll be fine with making the spec allowing for it :-)

sberyozkin:
@dblevins, I've updated my [to do list](https://github.com/eclipse/microprofile-jwt-auth/pull/149#issuecomment-598691007) to consider making it optional, I'll discuss it with other implementers in a couple of months when we'll be getting closer to 2.0, but as I said I won't mind myself.

And... good to talk to you again :-)

sberyozkin:
@dblevins Sorry, I've spent half a day at implementing the minimum JWE server support, got confused, but this API is pretty straghtforward to implement, certainly not longer than few days or not much longer. It took me awhile to implement it because I was experimenting, but I honestly don't expect anyone spend much time on implementing this API

sberyozkin:
@rdebusscher Hi Rudy, I think I'll make a summary of our disccussion with you and David and the pending review comments in the linked issue and will open a new fresh PR otherwise it will be difficult to continue a good discussion with so many comments already available. I'll do for MP JWT 2.0.

PR #153, opened 2020-02-27
Migrate TCK code to Jose4j, clean up and improve TokenUtils tests

### LINKED ISSUE #72: Could the TCK use auth0 instead of Nimbus JWT?
While there is nothing wrong with Nimbus Jose JWT, it is quite intrusive bringing its own JSON processing library (minidev). The KumuluzEE implementation uses auth0 java-jwt and e.g.:
[JWTPrincipal](https://github.com/kumuluz/kumuluzee-jwt-auth/blob/master/src/main/java/com/kumuluz/ee/jwt/auth/principal/JWTPrincipal.java) demonstrates, Auth0 works perfectly fine with the JSON-P instead of its own library.

sberyozkin:
Fixes #72.

This PR does the following:
- Updates `TokenUtils` which has become leaner
- Renamed `DebugTest` to `TokenUtilsGenerateTokenTest`, cleaned it up, added the checks the original token claims have made it into the verified form, added an invalid 'alg' test.
- Renamed `TokenUtilsTest` to `TokenUtilsMiscMethodsTest`, removed 2 of them since there were just printing something, with `TokenUtilsGenerateTokenTest` covering well the parsing/key loading, and updated 3 remaining tests to check the key length, and added the one to check a soon to be deprecated `1024` bit length can be created as well.

As I've mentioned at #72, even though it is the library we use in our own projects, it is really a good, compact and trusted library, so IMHO it will do well for the TCK


tck/src/test/java/org/eclipse/microprofile/jwt/tck/util/TokenUtilsGenerateTokenTest.java
```diff
@@ -0,0 +1,152 @@
...
+import org.jose4j.jwt.consumer.InvalidJwtException;
+import org.jose4j.jwt.consumer.JwtConsumerBuilder;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+/**
+ * Internal debugging tests
+ */
+public class TokenUtilsGenerateTokenTest {
+
```
rdebusscher: Within `DebugTest`, there was a `testHS256` test which doesn't seems to be retained. What is the reason for this?

tck/src/test/java/org/eclipse/microprofile/jwt/tck/util/TokenUtilsGenerateTokenTest.java
```diff
@@ -0,0 +1,152 @@
...
+    @Test(groups = TCKConstants.TEST_GROUP_UTILS, expectedExceptions = {InvalidJwtException.class},
+            description = "Illustrate validation of iss")
+    public void testFailAlgorithn() throws Exception {
```
rdebusscher: typo testFailAlgorith**n**

tck/src/test/java/org/eclipse/microprofile/jwt/tck/util/TokenUtilsGenerateTokenTest.java
```diff
@@ -0,0 +1,152 @@
...
+        // issuer must be equal to TCKConstants.TEST_ISSUER
+        builder.setExpectedIssuer(true, TCKConstants.TEST_ISSUER);
+        builder.setVerificationKey(publicKey);
+        builder.setAllowedClockSkewInSeconds(expGracePeriodSecs);
+
+        JwtClaims claimsSet = builder.build().processToClaims(token);
+        // Confirm all the claims available in /Token1.json have made it into the verified claimSet
+
+        Assert.assertEquals(claimsSet.getClaimsMap().size(), 18);
+        Assert.assertEquals("https://server.example.com", claimsSet.getIssuer());
```
rdebusscher: The signature is `Assert.assertEquals(Actual, Expected)` and thus parameters are swapped (also for the following lines)

tck/src/main/java/org/eclipse/microprofile/jwt/tck/util/TokenUtils.java
```diff
@@ -120,16 +112,19 @@ public static String generateTokenString(PrivateKey pk, String kid, String jsonR
             invalidClaims = Collections.emptySet();
         }
         InputStream contentIS = TokenUtils.class.getResourceAsStream(jsonResName);
+        if (contentIS == null) {
+            throw new IllegalStateException("Failed to find resource: " + jsonResName);
+        }
         byte[] tmp = new byte[4096];
```
rdebusscher: This is a much easier to read way of getting the content of the InputStream in a String

> Scanner s = new Scanner(inputStream).useDelimiter("\\A");
> String result = s.hasNext() ? s.next() : "";

tck/src/main/java/org/eclipse/microprofile/jwt/tck/util/TokenUtils.java
```diff
@@ -120,16 +112,19 @@ public static String generateTokenString(PrivateKey pk, String kid, String jsonR
             invalidClaims = Collections.emptySet();
         }
         InputStream contentIS = TokenUtils.class.getResourceAsStream(jsonResName);
```
rdebusscher: InputStream is never closed.

tck/src/test/java/org/eclipse/microprofile/jwt/tck/util/TokenUtilsMiscMethodsTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+/**
+ * Validation of the TokenUtils methods
+ */
+public class TokenUtilsMiscMethodsTest {
+
+    @Test(groups = TEST_GROUP_UTILS, description = "Used to generate a 2048 bit length key testing pair")
+    public void testKeyPairGeneration2048Length() throws Exception {
```
rdebusscher: Not very useful as this just test the JDK generation of an RSA key.

tck/src/test/java/org/eclipse/microprofile/jwt/tck/util/TokenUtilsMiscMethodsTest.java
```diff
@@ -0,0 +1,65 @@
...
+        KeyPair keyPair = TokenUtils.generateKeyPair(2048);
+        RSAPrivateKey privateKey = (RSAPrivateKey)keyPair.getPrivate();
+        Assert.assertEquals(privateKey.getModulus().bitLength(), 2048);
+        RSAPublicKey publicKey = (RSAPublicKey)keyPair.getPublic();
+        Assert.assertEquals(publicKey.getModulus().bitLength(), 2048);
+    }
+
+    @Test(groups = TEST_GROUP_UTILS, description = "Used to generate a 1024 bit length key testing pair")
+    public void testKeyPairGeneration1024Length() throws Exception {
```
rdebusscher: Not very useful as this just test the JDK generation of an RSA key.

tck/src/test/java/org/eclipse/microprofile/jwt/tck/util/TokenUtilsGenerateTokenTest.java
sberyozkin: @rdebusscher yes, I was looking at it, it was just testing that Nimbus could use `HS256` without even using `TokenUtils`. And we won't support the symmetric signatures for 1.2. So I dropped it but as mentioned I've introduced an invalid `alg` test which actually stresses the relevant TokenUtils code

tck/src/test/java/org/eclipse/microprofile/jwt/tck/util/TokenUtilsMiscMethodsTest.java
sberyozkin: Yeah, I was tempted to remove altogether but well, it is testing that `TokenUtils` is producing it correctly

sberyozkin:
@rdebusscher thanks, re the years, which years should we do, should we have a separate issue to change them across all the source in time for 1.2 ?

rdebusscher:
@sberyozkin  Regarding the years

Every new file which is added needs to have the year itself, thus 2020 now.

For every file which is updated, we need to update the license year to also include 2020 (update end year of the already present fork or add 2020 as end year).

No mass update of all files, only changed files require and update (and thus should be done in each PR)

sberyozkin:
@rdebusscher Thanks for the info...

tck/src/test/java/org/eclipse/microprofile/jwt/tck/util/TokenUtilsMiscMethodsTest.java
sberyozkin: The other thing this test is testing is that `2048` is enforced, same for the other test

sberyozkin:
Hi @rdebusscher I think I've addressed the comments, have a look please. I've set the year to 2020 in all the files, kept the lower margin year as these are not really new files, just renamed the test files.
Resolved few comments; those RSA key tests, yeah, they just check the utility methods do what they are expected to do. The main testing is in the generate token test method...

sberyozkin:
@rdebusscher Thanks

PR #155, opened 2020-02-28
Add methods to JsonWebToken to get claims by using the Claims enum

andreas-eberle:
fixes #154

PR #156, opened 2020-02-28
Revert "Add methods to JsonWebToken to get claims by using the Claims enum"

sberyozkin:
Reverts eclipse/microprofile-jwt-auth#155 due to this build failure:

https://ci.eclipse.org/microprofile/job/JWT-Auth-maven-snapshots/1063/

```
[ERROR] Failed to execute goal biz.aQute.bnd:bnd-baseline-maven-plugin:3.5.0:baseline (baseline) on project microprofile-jwt-auth-api: An error occurred while calculating the baseline: The baselining plugin detected versioning errors -> [Help 1]
```

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
+++ b/api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
@@ -48,7 +48,7 @@ public interface JsonWebToken extends Principal {
      * @return raw bear token string
      */
     default String getRawToken() {
-        return getClaim(Claims.raw_token);
+        return getClaim(Claims.raw_token.name());
     }
 
     /**
@@ -56,7 +56,7 @@ public interface JsonWebToken extends Principal {
      * @return the iss claim.
      */
     default String getIssuer() {
-        return getClaim(Claims.iss);
+        return getClaim(Claims.iss.name());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


sberyozkin:
CC @andreas-eberle

sberyozkin:
We don't have to revert if there is an accepted way to relax the baseline plugin policy

rdebusscher:
@sberyozkin there is nothing wrong with the PR so we need to fix baseline plugin policy.

Otherwise, every merge will fail.

sberyozkin:
@rdebusscher No, not every merge is failing, it started failing with this PR:
https://github.com/eclipse/microprofile-jwt-auth/commits/master

There were a few merges after @Emily-Jiang's PR was merged.
So, if you know how to fix it, please provide a PR but for now I need to revert it and reopen the issue as the build is broiken

rdebusscher:
@sberyozkin 

 Adding a default method is a binary compatible change. So the plugin is wrong here.

Also the output is wrong

```
[INFO] The baseline version was found to be 1.1.1
[ERROR] Baseline mismatch for package org.eclipse.microprofile.jwt, MINOR change. Current is 1.0.0, repo is 1.0.0, suggest 1.1.0 or -

```

So it correctly detects that current version in maven central is 1.1.1 and that there is a MINOR change required for the changes. But it fails the accept 1.2 as a MINOR change of 1.1.1 and suggests a lower version 1.1.0!!

And I even tried to play with a few other versions of the project. All of them are rejected!

- 2.0-SNAPSHOT
- 1.1-SNAPSHOT
- 1.2.0-SNAPSHOT
etc

So instead of reverting the PR in panic, one should look more detailed to the log and try to found the real cause.

And this is the first change in the API module after the baseline plugin is added to the pom.xml.

So as I mentioned, currently we can't merge any PR which touches the API module.

PR #158, opened 2020-03-03
Add get claims with claims, update bnd plugin version and increase jwt package osgi version to 1.1

### LINKED ISSUE #154: Allow retrieving claims by using the Claims enum
I think it would be a good idea to add a method `JsonWebToken.getClaim(Claims claim)` to allow retrieving claims by using the `Claims` enum. At the moment one is forced to write unnecessarily bloated code like `jwt.getClaim(Claims.groups.name())`. This is even true for the default methods given in the interface `JsonWebToken`. Adding this method could make code more readable.

Let me know if you want a PR for this.

andreas-eberle:
As discussed in #154 as well as #155 and #156.

This PR adds a default method to retrieve claims with the `Claims` enum. Furthermore, following the discussions mentioned above, the bnd plugin version is increased and the osgi package version of the jwt package is increased to 1.1.

fixes #154

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -28,6 +28,10 @@
     <name>MicroProfile JWT Auth API</name>
     <description>Eclipse MicroProfile JWT Feature - API</description>
 
+    <properties>
+        <bnd.version>5.0.0</bnd.version>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>jakarta.json</groupId>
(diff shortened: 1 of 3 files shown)


sberyozkin:
@andreas-eberle Thanks, does `mvn clean install` (with a `clean` task) passes for you locally ?

andreas-eberle:
Yes, doing a `mvn clean install` with Java8 in the top-level directory does finish successfully.

PR #164, opened 2020-03-16
Rename microprofile-config.properties to microprofile-config-publickey.properties

### LINKED ISSUE #135: Don't put `microprofile-config.properties` in META-INF
Currently, there is a `microprofile-config.properties` file present within META-INF of the test module for the TCK.

This can cause issues as it's a valid location to be picked up on the classpath, and the test jar is needed to run the tests. It would be better if the file was in a different location and then added to the tests that require it present in the deployment

sberyozkin:
Fixes #135 

Ken, I've only found one test using these properties, others have them created. Does it look OK ?

CC @starksm64 @rdebusscher, others


sberyozkin:
@radcortez Thanks, sorry missed it. Yeah let me go ahead with this one, even I can't break anything with this fix :-)

PR #165, opened 2020-03-16
Fix Readme

### LINKED ISSUE #152: update readme of repository
The readme says it is 'a proposal' in the Status section.

With JWT 1.1 out and preparing for JWT 1.2, I would say this is no longer a proposal but something real.

sberyozkin:
Fixes #152

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+// Copyright (c) 2016-2020 Contributors to the Eclipse Foundation
 //
 // See the NOTICES file(s) distributed with this work for additional
 // information regarding copyright ownership.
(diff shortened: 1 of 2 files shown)


README.adoc
```diff
@@ -19,17 +19,13 @@
...
+https://tools.ietf.org/html/rfc7519[JSON Web Tokens(JWT)] can be used for role based access control(RBAC) of microservice endpoints.
 
 
 ## Motivation
-MicroProfile 1.1 is a baseline platform definition that optimizes Enterprise Java for a microservices architecture and delivers application portability across multiple MicroProfile runtimes. While Java EE is a very feature rich platform and is like a toolbox that can be used to to address a wide variety of application architectures, MicroProfile focuses on defining a small and a minimum set of Java EE standards that can be used to deliver applications based on a microservice architecture, they are:
+MicroProfile 1.2 is a baseline platform definition that optimizes Enterprise Java for a microservices architecture and delivers application portability across multiple MicroProfile runtimes. While Java EE is a very feature rich platform and is like a toolbox that can be used to to address a wide variety of application architectures, MicroProfile focuses on defining a small and a minimum set of Java EE standards that can be used to deliver applications based on a microservice architecture, they are:
```
rdebusscher: Should it be MP 1.3 as we use Config 1.2 in the TCK?

sberyozkin: Sorry, I thought it was the upcoming MP JWT version that has to be released.
Should the version then be dropped from the README ? I don't mind fixing it to `1.3` but it won't change the definition of MP which follows. Let me know what you prefer, I'm fine either way

rdebusscher: The whole sentence/section can be removed as it is about MP, not about MP JWT.

sberyozkin: @rdebusscher Makes sense, how does it look now ?
I can look later on providing some more intro to JWT, add some recommendations about using OIDC when possible, etc...

rdebusscher: The section about why JWT tokens are used is still relevant and related to this project. So that can stay for me

> For RESTful based microservices, security tokens offer a very lightweight and interoperable way to propagate identities across different services, where:
> 
> * Services don’t need to store any state about clients or users
> * Services can verify the token validity if token follows a well known format. Otherwise, services may invoke a separated service.
> * Services can identify the caller by introspecting the token. If the token follows a well known format, services are capable to introspect the token by
> themselves, locally. Otherwise, services may invoke a separated service.
> * Services can enforce authorization policies based on any information within a security token
> * Support for both delegation and impersonation of identities

sberyozkin: @rdebusscher Oh, I haven't paid the attention, thought it was all about MP only, will it bring this fragment back

sberyozkin: @rdebusscher I've restored it with the few minor updates, have a look please

sberyozkin:
@rdebusscher thanks

PR #168, opened 2020-03-19
Fix All JAXRS TCK tests and update the licence year for the affected tests

### LINKED ISSUE #135: Don't put `microprofile-config.properties` in META-INF

sberyozkin:
This PR completes the #164 PR.
Fixes #135

### DIFF
--- a/tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/InvalidTokenTest.java
+++ b/tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/InvalidTokenTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016-2020 Contributors to the Eclipse Foundation
  *
  *  See the NOTICE file(s) distributed with this work for additional
  *  information regarding copyright ownership.
@@ -60,6 +60,7 @@ public class InvalidTokenTest extends Arquillian {
      */
     @Deployment(testable=true)
     public static WebArchive createDeployment() throws IOException {
+        URL config = InvalidTokenTest.class.getResource("/META-INF/microprofile-config-publickey-location.properties");
         URL publicKey = InvalidTokenTest.class.getResource("/publicKey.pem");
         WebArchive webArchive = ShrinkWrap
             .create(WebArchive.class, "InvalidTokenTest.war")
@@ -68,7 +69,7 @@ public class InvalidTokenTest extends Arquillian {
             .addClass(RolesEndpoint.class)
             .addClass(TCKApplication.class)
             .addAsWebInfResource("beans.xml", "beans.xml")
-            ;
+            .addAsManifestResource(config, "microprofile-config.properties");
         System.out.printf("WebArchive: %s\n", webArchive.toString(true));
         return webArchive;
     }
--- a/tck/README.adoc
+++ b/tck/README.adoc
@@ -49,6 +49,8 @@ invalid tokens. A summary of the tck/resources directory contents is:
 
 * META-INF/microprofile-config-publickey.properties
 ** A base MicroProfile config properties file for the org.eclipse.microprofile.jwt.tck.config.* package tests.
+* META-INF/microprofile-config-publickey-location.properties
+** A base MicroProfile config properties file for the org.eclipse.microprofile.jwt.tck.jaxrs.* package tests.
 * jwt-content1.json
 * testJWTCallerPrincipal.json
 * usePreferredName.json
--- a/tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/PublicKeyAsPEMTest.java
+++ b/tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/PublicKeyAsPEMTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2018 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016-2020 Contributors to the Eclipse Foundation
  *
  *  See the NOTICE file(s) distributed with this work for additional
  *  information regarding copyright ownership.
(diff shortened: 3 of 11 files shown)


sberyozkin:
@radcortez, Hi Roberto, I've run Quarkus TCK with this update and I'm pretty sure it is all good now, this test's invalid issue test was failing because it was relying before on the issuer property set in those mciroprofile properties. Also updated the license year of the updated tests.

CC @rdebusscher

radcortez:
Looks good. Let me try to run this in SmallRye too.

We can now run the TCK there as well :)
https://github.com/smallrye/smallrye-jwt/pull/200

sberyozkin:
@radcortez Cool, thanks a million

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/InvalidTokenTest.java
```diff
@@ -60,6 +61,7 @@
      */
     @Deployment(testable=true)
     public static WebArchive createDeployment() throws IOException {
+        URL config = PublicKeyAsPEMTest.class.getResource("/META-INF/microprofile-config.properties");
```
radcortez: This needs to be "/META-INF/microprofile-config-publickey.properties".

radcortez:
I think there are other tests that we need to explicitly add the microprofile config file (because they were relying on the default location):

- ClaimValueInjectionTest
- JsonValueInjectionTest
- PrimitiveInjectionTest
- ProviderInjectionTest
- RequiredClaimsTest
- RolesAllowedTest

I can do that.

sberyozkin:
@radcortez I did verify that on the jwe branch, then switched and copied the wrong code, but I'm 100% verified the Quarkus build...

radcortez:
Sure. I've just found it by running in my local TCK branch :)

sberyozkin:
@radcortez Cool, let me complete it then, thanks for offering to help

sberyozkin:
@radcortez Figured out how I ended up copying that wrong code, I forgot to rebase and assumed it was already correct when copying from PublickeyAsPEMTest, glad you have checked :-)

radcortez:
Cool :)

sberyozkin:
@radcortez I've cleaned it up and also removed few unused properties, double check please when you have some time :-)

Rudy @rdebusscher - any concerns, ping us please

radcortez:
Hum... I'm getting a 401 running the tests now.

Shouldn't `microprofile-config-publickey.properties` have the content of `publicly.pem` in `mp.jwt.verify.publickey`? It doesn't seem to match.

The tests use the default `privateKey.pem` to sign the token.

radcortez:
And we missed `PrincipalInjectionTest`.

sberyozkin:
@radcortez I'll investigate a bit later, it probably means that those tests did not need those properties then. I'll start from scratch...Thanks

radcortez:
Yes, I thought about that as well, but if we had the `microprofile-config.properties` before, it should have been using that by default.

If you look into Wildfly TCK they do setup a `microprofile-config.properties` if there is none:
https://github.com/wildfly/wildfly/blob/9160a2ca507b909b818ccc128291a5468b83b11a/testsuite/integration/microprofile-tck/jwt/src/test/java/org/wildfly/extension/microprofile/jwt/test/DeploymentProcessor.java#L43

And the file has the `publickKey-pem`:
https://github.com/wildfly/wildfly/blob/9160a2ca507b909b818ccc128291a5468b83b11a/testsuite/integration/microprofile-tck/jwt/src/test/resources/META-INF/microprofile-config-local.properties

I did the same for SmallRye JWT TCK.

sberyozkin:
Hi @radcortez I suspect these TCK tests are not actually set up properly and we have opened something bigger here :-).  While I was off on a walk (now I know why I have a terrible headache :-)) I realized that sll of these tests already add `publicKey.pem` and TCK Readme does say that this `microprofile-config.properties` (now renamed to `microprofile-config-publickey.properties`) is only for the config tests. But TCK META-INF folder also has a `microprofile-config-location.properties` or something like that and it is really that file which should be used here. 
I'll have a look tomorrow, I'll likely close this PR and restart afresh.
Thanks for your feedback, cheers

radcortez:
@sberyozkin I think I got it.

These tests are from MP 1.0, when we didn't have MP Config integration yet and keys and other configurations were provided by vendor specific ways. So the tests didn't have a way to provide the configuration.

When we introduced the MP Config integration, some tests started to use microprofile-config.properties, but the vendor specific way will always have priority over that, so even if the microprofile-config.properties was not right for the test, the previous configuration from the vendor would take over and pass the TCK (if they were already passing it for 1.0).

Does this make sense? I think we can change these tests to provide the right configuration and not rely on vendors to provide their way (or projects supplying their own microprofile-config.properties, like Wildfly or Quarkus).

sberyozkin:
Hi @radcortez 

That is very helpful, so yeah, I've just updated the branch locally and confirmed with the smallrye-jwt tck (I still can't suppress a happy smile the project finally has it :-), it won't guarantee in itself the integrators are compliant but it does offer a robust first barrier check), by commenting out the existing `microprofile-config.properties` there, that having the proper configuration shipped with the TCK itself, all works. So the implementers won't have to create local properties.
PR update follows shortly.
CC @rdebusscher, others

radcortez:
Cool :)

sberyozkin:
@radcortez Done. I've introduced `META-INF/microprofile-config-publickey-location.properties` and removed `META-INF/microprofile-config-location.properties` because it was not referenced in README, not used anywhere in TCK and had an issuer property mistyped, so pretty sure it can go...
Verified with smallrye-jwt that it works with the project specific `microprofile-config.properties` or without it - please verify independently.
Note I'm going to open another issue to update the optional tests too as well as fix the full tck suite file as it contains a fraction of the base tests only
CC @radcortez

radcortez:
Thanks! I'll have a look.

radcortez:
I've also tested with SmallRye and it does work with the base tests.

sberyozkin:
@radcortez Thanks, going to go ahead with the merge shortly

PR #172, opened 2020-03-23
Added Github Action Workflow for Build.

radcortez:
We could move to Github Actions to handle Build and Release process for MicroProfile Config. Some of the main motivations to move are:

- Run the build in forks. You don't need to do the PR to check the status of the build
- Change the Build / Release configuration by just submitting a PR
- Committers have control over the CI
- Flexible to execute actions on Github Events

I've detailed some of the work done over SmallRye to use Github Actions here:
https://github.com/smallrye/smallrye-parent/wiki/SmallRye-Release-Process

The idea is not to pull the plug on Jenkins yet. We can run both in parallel. We can start with just the build part, which is easy to implement and go from there.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,47 @@
+name: MicroProfile JWT Build
+
+on:
+  push:
+    paths-ignore:
+      - 'editorconfig'
+      - '.gitattributes'
+      - '.gitignore'
+      - 'CONTRIBUTING*'
+      - 'CODEOWNERS'
+      - 'KEYS'
+      - 'LICENSE'
+      - 'NOTICE'
+      - 'README*'
+      - 'site.yaml'
+  pull_request:
+    paths-ignore:
+      - 'editorconfig'
+      - '.gitattributes'
+      - '.gitignore'
+      - 'CONTRIBUTING*'
+      - 'CODEOWNERS'
+      - 'KEYS'
+      - 'LICENSE'
+      - 'NOTICE'
+      - 'README*'
+      - 'site.yaml'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [8, 11]
+    name: build with jdk ${{matrix.java}}
+
+    steps:
+      - uses: actions/checkout@v2
+        name: checkout
+
+      - uses: actions/setup-java@v1.3.0
+        name: set up jdk ${{matrix.java}}
+        with:
+          java-version: ${{matrix.java}}
+
+      - name: build with maven
+        run: mvn verify
--- a/pom.xml
+++ b/pom.xml
@@ -325,6 +325,7 @@
                         <exclude>.factorypath</exclude>
                         <exclude>.editorconfig</exclude>
                         <exclude>sandbox/**</exclude>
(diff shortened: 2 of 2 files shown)


radcortez:
This requires https://github.com/eclipse/microprofile-jwt-auth/pull/171 for the JDK 11 build.

sberyozkin:
Hey @radcortez, I wonder if we should create out MP JWT specific forum and post this proposal there.
And give it a few weeks. I'll open an issue to do it anyway...
Right now I'd rather deal with it post 1.2 as there is still a lot of work to do for it so I'd be happy with the existing process for now.

radcortez:
Yes.

Actually, I plan to propose this as a wide process change to all MicroProfile projects.

Maybe this description was a little off. This is the final goal, but this PR is just for the Build part. I see no harm in adding it right now, and could help us catch some issues. For instance, when I broke the build with the TCK suite files move :)

sberyozkin:
@radcortez Was there a broken build in that PR ?

sberyozkin:
Hmm, I was naively assuming the PRs were triggering a build :-), but the only checks are signature checks :-). So the only way to catch the failure early before the merge is to build locally ?

radcortez:
I think there is a Jenkins server at Eclipse (isn't it how you do releases?), but the build result is not being reported to Github. With the GH Action we would get it for free, plus it also works on forks.

With, right now you need to build locally. It was my mistake, you should do that anyway and not rely that the project is going to report that for you, but still I think it is convenient.

sberyozkin:
@radcortez FYI, MP JWT 1.2 will be the 1st release I'll be doing. That Jenkins server does report the issues if any but after we merge the PRs, when we had a PR which was causing the build failure after I merged it, I got the notification.
So what we really want to is when someone opens a PR a CI is started and checks it immediately. GH Actions will give it to us, this is something I get :-), lets do it

.github/workflows/build.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: actions/checkout@v2
+        name: checkout
+
+      - uses: actions/setup-java@v1.3.0
+        name: set up jdk ${{matrix.java}}
+        with:
+          java-version: ${{matrix.java}}
+
+      - name: build with maven
+        run: mvn verify
```
sberyozkin: @radcortez Hi Roberto, I believe that should include `clean` to get the osgi versioning plugin checking things

radcortez: You don't need it, because GH Actions build starts fresh every time, meaning empty Maven repo and no previous target folder.

radcortez:
Ah, so Jenkins is probably only setup to build on push to master and not on PR's.

Yes, this will build when someone opens a PR, so we get the feedback before merging :)

.github/workflows/build.yml
sberyozkin: I recall Emily saying something about `clean' related to that plugin, but sure, we can add if needed

radcortez: Most likely, because Jenkins keeps the state of the previous build and when you build again, you have the old build version there, but GH Actions just start everything in a new container.  On each build, there are no old binaries and even all the Maven Dependencies are downloaded again, so I think we don't need clean.

Issue #121, opened 2019-01-15
Add mpJwt audiences configuration property via mpConfig


chunlongliang-ibm:
The audiences is a common claim in JWT. If JWT contains "aud", the recipient is required to verify the audiences. Current mpconfig have configurations to verify issuer and signature only, need address other verifiable claims.

MikeEdgar:
+1

This relates to smallrye/smallrye-jwt#110 and loosely to #63. 

Suggested properties:
* Set/List of audiences -> `mp.jwt.verify.audiences` 
* Whether the `aud` claim is required -> `mp.jwt.verify.requireaud`

As mentioned in the first comment, for the second property, it may be better to instead introduce a general property to allow specification of a list of required claims.

MikeEdgar:
I would suggest this issue be expanded to generally support claim verification in an extensible way. There was a great suggestion in the MP JWT 1.1 draft discussion proposing the following as a way to verify a particular claim:

> Using a `mp.jwt.verify.claim.*` namespace could potentially allow us use the Config API to list the properties under the `mp.jwt.verify.claim.` and have all sub entries processed as claims using the format:
>
> - `mp.jwt.verify.claim.<claim> = <regex>`

And also a way to list the claims that are required:
> - `mp.jwt.verify.claims.required=sub,iss,color,groups,name,age`

Are there any thoughts on solving this audience configuration issue more broadly?

sberyozkin:
@MikeEdgar I propose we follow your original idea and add `mp.jwt.verify.audiences` (only without `s`) as it is a well known claim and introduce something more generic at some later stage.

 @chunlongliang-ibm Can you please create a PR ?

chunlongliang-ibm:
I would like to propose that we add mp-config property "mp.jwt.verify.audiences", and accept multiple audiences in comma delimited format, like
mp.jwt.verify.audiences=audience1, audience2, audience3

To maintain back-compatibility, there is no default value  for mp.jwt.verify.audiences. If mp.jwt.verify.audiences is not configured, and JWT contains "aud", it is up to service to interpret the default behavior.

sberyozkin:
@chunlongliang-ibm sounds good to me, can you please start working on a PR ? Thanks

[...]

PR #175, opened 2020-03-23
add mp.jwt.verify.audiences requirements and tests

brutif:
Signed-off-by: Bruce Tiffany <btiffany@us.ibm.com>

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/jwt/config/Names.java
+++ b/api/src/main/java/org/eclipse/microprofile/jwt/config/Names.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016-2020 Contributors to the Eclipse Foundation
  *
  *  See the NOTICE file(s) distributed with this work for additional
  *  information regarding copyright ownership.
@@ -54,4 +54,10 @@ public interface Names {
      * This configuration will be ignored unless `mp.jwt.token.header` is set to `Cookie`.
      */
     String TOKEN_COOKIE = "mp.jwt.token.cookie";
+    
+    /**
+     * The expected "aud" claim value(s), separated by commas. 
+     * If specified, MP-JWT claim must be present and match one of the values.
+     */
+    String AUDIENCES = "mp.jwt.verify.audiences";
 }
--- a/spec/src/main/asciidoc/configuration.asciidoc
+++ b/spec/src/main/asciidoc/configuration.asciidoc
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2016-2018 Eclipse Microprofile Contributors:
+// Copyright (c) 2016-2020 Eclipse Microprofile Contributors:
 // Red Hat, IBM, Tomitribe
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
@@ -356,6 +356,11 @@ return Public Key text in one of the supported formats.
 
 The `mp.jwt.verify.issuer` config property allows for the expected value of the `iss`
 claim to be specified.  A MicroProfile JWT implementation must verify the `iss` claim of incoming JWTs is present and matches the configured value of `mp.jwt.verify.issuer`.
+
+#### `mp.jwt.verify.audiences`
+
+The `mp.jwt.verify.audiences` config property is a comma delimited list of allowable values for the `aud` claim.  If specified, a MicroProfile JWT implementation must verify the `aud` claim of incoming JWTs is present and at least one value in the claim matches one of the configured values of `mp.jwt.verify.audiences`.
+Note that the aud claim should only be injected into a Set of String, not as a comma delimited string.
 
 ## JWT and HTTP headers
 
--- a/tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/AudValidationBadAudTest.java
+++ b/tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/AudValidationBadAudTest.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) 2016-2020 Contributors to the Eclipse Foundation
+ *
+ *  See the NOTICE file(s) distributed with this work for additional
+ *  information regarding copyright ownership.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  You may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package org.eclipse.microprofile.jwt.tck.container.jaxrs;
+
+import java.io.StringWriter;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.security.PrivateKey;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.HttpHeaders;
+import javax.ws.rs.core.Response;
+
+import org.eclipse.microprofile.jwt.config.Names;
+import org.eclipse.microprofile.jwt.tck.TCKConstants;
+import org.eclipse.microprofile.jwt.tck.config.SimpleTokenUtils;
+import org.eclipse.microprofile.jwt.tck.util.MpJwtTestVersion;
+import org.eclipse.microprofile.jwt.tck.util.TokenUtils;
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.test.api.ArquillianResource;
+import org.jboss.arquillian.testng.Arquillian;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.testng.Assert;
+import org.testng.Reporter;
+import org.testng.annotations.Test;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static org.eclipse.microprofile.jwt.tck.TCKConstants.TEST_GROUP_CONFIG;
+
+/**
+ * Validate the handling of the JWT aud claim.
+ *
+ * Validate the aud claim against the {@linkplain Names#AUDIENCES} property is performed, and fails
+ * if the jwt aud does not match the {@linkplain Names#AUDIENCES} property.
+ */
+public class AudValidationBadAudTest extends Arquillian {
+    /**
+     * The base URL for the container under test
+     */
+    @ArquillianResource
+    private URL baseURL;
+
+    /**
+     * The token used by the test
+     */
+    private static String token;
+
+    /**
+     * Create a CDI aware base web application archive that includes an embedded PEM public key
+     * that is included as the mp.jwt.verify.publickey property.
+     * The root url is /
+     * @return the base base web application archive
+     * @throws Exception - on resource failure
+     */
+    @Deployment()
+    public static WebArchive createDeployment() throws Exception {
+        URL publicKey = AudValidationBadAudTest.class.getResource("/publicKey4k.pem");
+
+        PrivateKey privateKey = TokenUtils.readPrivateKey("/privateKey4k.pem");
+        String kid = "publicKey4k";
+        Map<String, Long> timeClaims = new HashMap<>();
+        token = TokenUtils.generateTokenString(privateKey, kid, "/TokenAudiences.json", null, timeClaims);
+
+        // Setup the microprofile-config.properties content
+        Properties configProps = new Properties();
+        // Location points to the PEM bundled in the deployment
+        configProps.setProperty(Names.VERIFIER_PUBLIC_KEY_LOCATION, "/publicKey4k.pem");
+        configProps.setProperty(Names.ISSUER, TCKConstants.TEST_ISSUER);
+        configProps.setProperty(Names.AUDIENCES, "notAValidAudience");  // does not match json, should fail
+        StringWriter configSW = new StringWriter();
+        configProps.store(configSW, "AudValidationBadAudTest microprofile-config.properties");
+        StringAsset configAsset = new StringAsset(configSW.toString());
+
+        WebArchive webArchive = ShrinkWrap
+                .create(WebArchive.class, "AudValidationBadAudTest.war")
+                .addAsManifestResource(new StringAsset(MpJwtTestVersion.MPJWT_V_1_2.name()), MpJwtTestVersion.MANIFEST_NAME)
+                .addAsResource(publicKey, "/publicKey.pem")
+                .addAsResource(publicKey, "/publicKey4k.pem")
+                // Include the token for inspection by ApplicationArchiveProcessor
+                .add(new StringAsset(token), "MP-JWT")
+                .addClass(AudienceValidationEndpoint.class)
+                .addClass(TCKApplication.class)
+                .addClass(SimpleTokenUtils.class)
+                .addAsWebInfResource("beans.xml", "beans.xml")
+                .addAsManifestResource(configAsset, "microprofile-config.properties");
+        System.out.printf("WebArchive: %s\n", webArchive.toString(true));
+        return webArchive;
+    }
+
+    @RunAsClient
+    @Test(groups = TEST_GROUP_CONFIG,
+        description = "Validate that JWK with aud that is not contained in mp.jwt.verify.audiences returns HTTP_UNAUTHORIZED")
+    public void testRequiredAudMismatchFailure() throws Exception {
+        Reporter.log("testRequiredAudMismatchFailure, expect HTTP_UNAUTHORIZED");
+
+        String uri = baseURL.toExternalForm() + "endp/verifyAudIsOk";
+        WebTarget echoEndpointTarget = ClientBuilder.newClient()
+            .target(uri)
+            ;
+        Response response = echoEndpointTarget.request(APPLICATION_JSON).header(HttpHeaders.AUTHORIZATION, "Bearer "+token).get();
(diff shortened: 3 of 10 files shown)


brutif:
addresses #121

sberyozkin:
@brutif thanks, nice work, can you please review the comments I left at #174 ?

brutif:
@sberyozkin will do

api/src/main/java/org/eclipse/microprofile/jwt/config/Names.java
```diff
@@ -40,4 +40,10 @@
      * The expected iss claim value to validate against an MP-JWT.
      */
     String ISSUER = "mp.jwt.verify.issuer";
+    
+    /**
+     * The expected "aud" claim value(s), separated by commas.
```
rdebusscher: This is not exactly the same as in the spec doc. Here it appears that "aud" claim value must match exactly with the contents of the value of the property.

To avoid confusing, harmonize the texts.

spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -357,6 +357,11 @@ return Public Key text in one of the supported formats.
 The `mp.jwt.verify.issuer` config property allows for the expected value of the `iss`
 claim to be specified.  A MicroProfile JWT implementation must verify the `iss` claim of incoming JWTs is present and matches the configured value of `mp.jwt.verify.issuer`.
 
+#### `mp.jwt.verify.audiences`
+
+The `mp.jwt.verify.audiences` config property is a comma delimited list of allowable values for the `aud` claim.  If specified, a MicroProfile JWT implementation must verify the `aud` claim of incoming JWTs is present and at least one value in the claim matches one of the configured values of `mp.jwt.verify.audiences`.
+If the JWT contains an `aud` claim and the `mp.jwt.verify.audiences` config property is not set, behavior is undefined for backward compatibility purposes.
```
rdebusscher: The backwards compatible behaviour is that JWT is allowed which is not the same as undefined behavior. Please adapt the wording.

spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -357,6 +357,11 @@ return Public Key text in one of the supported formats.
 The `mp.jwt.verify.issuer` config property allows for the expected value of the `iss`
 claim to be specified.  A MicroProfile JWT implementation must verify the `iss` claim of incoming JWTs is present and matches the configured value of `mp.jwt.verify.issuer`.
 
+#### `mp.jwt.verify.audiences`
+
+The `mp.jwt.verify.audiences` config property is a comma delimited list of allowable values for the `aud` claim.  If specified, a MicroProfile JWT implementation must verify the `aud` claim of incoming JWTs is present and at least one value in the claim matches one of the configured values of `mp.jwt.verify.audiences`.
```
rdebusscher: Probably should mention that order in the config property is not of important.

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/PublicKeyEndpoint.java
```diff
@@ -559,6 +568,55 @@ else if(issuer.isPresent()) {
...
+    @Path("/verifyAudIsOk")
+    @Produces(MediaType.APPLICATION_JSON)
+    @RolesAllowed("Tester")
+    public JsonObject verifyAudIsOk() {
+        boolean pass = false;
+        String msg;
+
+        if(!aud.getValue().isPresent()) {
+            // The iss claim should be provided for this endpoint
+            msg = String.format("MP-JWT missing aud claim, or injection of claim failed");
```
rdebusscher: String.format without placeholder. Can be simple String.

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/PublicKeyEndpoint.java
```diff
@@ -559,6 +568,55 @@ else if(issuer.isPresent()) {
...
+        }
+        else if(audiences.isPresent()) {
+            
+            Set<String> claimAud = aud.getValue().get();
+            String[] configAud = audiences.get().split(",");
+            boolean match = false;
+            Iterator<String> it = claimAud.iterator();
+            while(it.hasNext()) {
```
rdebusscher: Suggestions: enhanced for loop for while and for loop at line 596.

spec/src/main/asciidoc/configuration.asciidoc
sberyozkin: `matches one of the configured values` already gives it

spec/src/main/asciidoc/configuration.asciidoc
sberyozkin: I've proposed to drop this sentence altogether at #174 as it only raises question. The property is optional hence there is no need to say what happens if this property is not set

tck/src/test/resources/Token1.json
```diff
@@ -3,8 +3,8 @@
     "jti": "a-123",
     "sub": "24400320",
     "upn": "jdoe@example.com",
-    "preferred_username": "jdoe",
-    "aud": "s6BhdRkqt3",
+    "preferred_username": "jdoe",    
+    "aud": ["s6BhdRkqt3","aud1", "aud2", "aud3", "aud4"],
```
sberyozkin: Please remove some extra spaces here

tck/src/test/resources/Token1.json
```diff
@@ -3,8 +3,8 @@
     "jti": "a-123",
     "sub": "24400320",
     "upn": "jdoe@example.com",
-    "preferred_username": "jdoe",
-    "aud": "s6BhdRkqt3",
+    "preferred_username": "jdoe",
```
sberyozkin: Extra space in the end of `preferred_username` and just before `custom_string`

tck/src/test/resources/Token1.json
sberyozkin: `aud` can be both a string containing a single value and an array as in this PR. We should keep `Token1.json` resource unchanged and introduce `TokenAudiences.json` where `aud` is this array.

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/AudValidationTest.java
```diff
@@ -0,0 +1,136 @@
...
+        String kid = "publicKey4k";
+        HashMap<String, Long> timeClaims = new HashMap<>();
+        token = TokenUtils.generateTokenString(privateKey, kid, "/Token1.json", null, timeClaims);
+
+        // Setup the microprofile-config.properties content
+        Properties configProps = new Properties();
+        // Location points to the PEM bundled in the deployment
+        configProps.setProperty(Names.VERIFIER_PUBLIC_KEY_LOCATION, "/publicKey4k.pem");
+        configProps.setProperty(Names.ISSUER, TCKConstants.TEST_ISSUER);
+        configProps.setProperty(Names.AUDIENCES, "aud3");  // matches json, should pass
```
sberyozkin: Use more than one value here

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/AudValidationTest.java
```diff
@@ -0,0 +1,136 @@
...
+     * @throws Exception - on resource failure
+     */
+    @Deployment()
+    public static WebArchive createDeployment() throws Exception {
+        URL publicKey = AudValidationTest.class.getResource("/publicKey4k.pem");
+
+        PrivateKey privateKey = TokenUtils.readPrivateKey("/privateKey4k.pem");
+        String kid = "publicKey4k";
+        HashMap<String, Long> timeClaims = new HashMap<>();
+        token = TokenUtils.generateTokenString(privateKey, kid, "/Token1.json", null, timeClaims);
```
sberyozkin: Keep this test which uses `/Token1.json` and have a duplicate which will use `/TokenAudiences.json` (see below)

spec/src/main/asciidoc/configuration.asciidoc
brutif: dropped per 174

spec/src/main/asciidoc/configuration.asciidoc
sberyozkin: Rudy, @rdebusscher, resolving this one, please reopen if disagree

api/src/main/java/org/eclipse/microprofile/jwt/config/Names.java
brutif: fixed, I think

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/PublicKeyEndpoint.java
brutif: fixed, thanks

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/PublicKeyEndpoint.java
brutif: good idea, thanks

tck/src/test/resources/Token1.json
brutif: fixed

tck/src/test/resources/Token1.json
brutif: done

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/AudValidationTest.java
brutif: done

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/AudValidationTest.java
brutif: missed this, fixing now

brutif: fixed

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/AudValidationTest2.java
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Validate the handling of the JWT aud claim.
+ *
+ * Validate the aud claim against the {@linkplain Names#AUDIENCES} property is performed, and passes
+ * if the jwt aud is contained in the {@linkplain Names#AUDIENCES} property.
+ * 
+ * This test checks against a claim containing multiple values.
+ * 
+ */
+public class AudValidationTest2 extends Arquillian {
```
sberyozkin: @brutif Can you please rename it to `AudArrayVaidationTest` - would likely be easier to understand what it is about

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/PublicKeyEndpoint.java
```diff
@@ -32,6 +32,7 @@
 import java.security.interfaces.RSAPublicKey;
 import java.util.Base64;
 import java.util.Optional;
+import java.util.Set;
```
sberyozkin: @brutif As suggested at #174 please also create a `jaxrs/AudienceValidationEndpoint` as there is a pattern there in `jaxrs` that every test has a dedicated endpoint. This `AudienceValidationEndpoint` would only have those audiences injected.

api/src/main/java/org/eclipse/microprofile/jwt/config/Names.java
sberyozkin: @rdebusscher resolving, LGTM, please reopen if you feel an extra tweak is needed

sberyozkin: @rdebusscher LGTM, reopen if needed

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/PublicKeyEndpoint.java
sberyozkin: Oops, you've already do it :-), but please revert the injection code in `config/PublickeyEndpoint` :-) as it is not used anywhere there

sberyozkin:
@brutif It looks very good after addressing @rdebusscher 's comments and moving the tests, I'm asking for two more minor updates, and on top of it, please git fetch and rebase against `upstream/master`, and squash all your commits and force-push.

sberyozkin:
We also need to specify what happens when an `aud` claim is an array and the code is:
```
    @Inject
    @Claim(standard = Claims.aud)
    private ClaimValue<Optional<String>> aud;
```

I don't think the spec should say that an array needs to be flattened into a comma separated string. Instead the spec should something like this with the better wording:
```
Audience claim value should be injected to a Set of strings because an audience claim may be an array of strings. If the audience claim is an array and is injected into a String then no guarantee can be made which member of the audience claim array will be injected into this String.
```
If we don't say that immediately after 1.2 will be released there will be some questions asked so it is better just to clarify it now (and in the future generalize this guideline to the other claim array values)

sberyozkin:
@brutif And please add the new `Aud*` TCK tests to `test/resources/suites/tck-basic-suite.xml` (I can deal with `tck-full-suite.xml` later as part of another open issue)

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/AudValidationBadAudTest.java
```diff
@@ -0,0 +1,127 @@
...
+     * @return the base base web application archive
+     * @throws Exception - on resource failure
+     */
+    @Deployment()
+    public static WebArchive createDeployment() throws Exception {
+        URL publicKey = AudValidationBadAudTest.class.getResource("/publicKey4k.pem");
+
+        PrivateKey privateKey = TokenUtils.readPrivateKey("/privateKey4k.pem");
+        String kid = "publicKey4k";
+        HashMap<String, Long> timeClaims = new HashMap<>();
```
sberyozkin: This is a minor one, but I've been keeping updating other TCK source :-), so please have `Map<...>` here and other new tests

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/AudValidationBadAudTest.java
```diff
@@ -0,0 +1,127 @@
...
+        configProps.setProperty(Names.VERIFIER_PUBLIC_KEY_LOCATION, "/publicKey4k.pem");
+        configProps.setProperty(Names.ISSUER, TCKConstants.TEST_ISSUER);
+        configProps.setProperty(Names.AUDIENCES, "notAValidAudience");  // does not match json, should fail
+        StringWriter configSW = new StringWriter();
+        configProps.store(configSW, "AudValidationBadAudTest microprofile-config.properties");
+        StringAsset configAsset = new StringAsset(configSW.toString());
+
+        WebArchive webArchive = ShrinkWrap
+                .create(WebArchive.class, "AudValidationBadAudTest.war")
+                .addAsManifestResource(new StringAsset(MpJwtTestVersion.MPJWT_V_1_1.name()), MpJwtTestVersion.MANIFEST_NAME)
```
sberyozkin: Just realized this has to be `MpJwtTestVersion.MPJWT_V_1_2` (and in other new tests)

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/AudValidationTest2.java
brutif: done

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/PublicKeyEndpoint.java
brutif: done

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/AudValidationBadAudTest.java
brutif: done

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/AudValidationBadAudTest.java
brutif: fixed

brutif:
@sberyozkin Regarding injecting a string for the value of aud, should that even work? Claims.java says it needs to be a set. "For every claim in this enum, an MP-JWT implementation must return a value of the indicated type".   There's an older test, RequiredClaimsEndpoint:178, that uses a set.   As far as the commas go, configuration.asciidoc states those are used in the value of the config property, but not in the value of the claim, so I think we may be ok?  If there's some other wording that would make it more clear, I can include that when I squash.

sberyozkin:
@brutif Thanks, please push the changes once you are finished and also keep squashing the commits

sberyozkin:
> Regarding injecting a string for the value of aud, should that even work? Claims.java says it needs to be a set. "For every claim in this enum, an MP-JWT implementation must return a value of the indicated type"

Good point, I've missed it.

> If there's some other wording that would make it more clear, I can include that when I squash.

I missed it and may be I won't be alone in missing it :-), and someone will try to inject into `String`.
So perhaps:
`Note that an audience `aud` claim should only be injected into Set to avoid losing some audience values` or something better along these lines if you prefer.

Please update, squash and hopefully this PR will be ready to go

api/src/main/java/org/eclipse/microprofile/jwt/config/Names.java
```diff
@@ -40,7 +40,7 @@
      * The expected iss claim value to validate against an MP-JWT.
      */
     String ISSUER = "mp.jwt.verify.issuer";
-
+
```
sberyozkin, line 43: @brutif Extra space here

sberyozkin:
Hi @brutif I have only left one minor comment.
@rdebusscher Does it look OK to you now ?

api/src/main/java/org/eclipse/microprofile/jwt/config/Names.java
brutif: fixed

sberyozkin:
@rdebusscher I'm now ready to merge - please confirm your change request can be lifted

sberyozkin:
@rdebusscher I don't really follow what you mean about a breaking change since `Claims` documentation which I have become aware makes it obvious that only `Set` must be used. However I agree enforcing it now could indeed break some user code as I suppose the implementations (ours including) may be a bit lax about it.

So yes, lets have this hint and consider enforcing it later I suppose.

sberyozkin:
@rdebusscher I've looked around and I see no dedicated section dedicated to the injection of various claims, so for now this is this is the best place to make this hint visible. Once we sort the typed injection of all sort of claims then a dedicated section can be created

Issue #93, opened 2018-05-17
Allow for custom authorization header and scheme


jeanouii:
It is very common to use `Authorization` header with `Bearer` scheme.
But there must be more than one JWT and therefor it would be useful to make this configurable even if they default to `Authorization` header and `Bearer` scheme.

As discussed together, this won't go into this release but we should address this issue later.

chunlongliang-ibm:
Many existing applications have header 
Authorization: Token <JWT here>
For example,  https://github.com/gothinkster/realworld/tree/master/api

sberyozkin:
We have 3 properties in `smallrye-jwt` to deal with it:
`smallrye.jwt.token.header`
`smallrye.jwt.token.cookie`
`smallrye.jwt.token.scheme`

The first 2 headers support non `Authorization` headers, the 2nd one - a cookie name if the 1st property is set to `Cookie` and the 3rd one - a custom `Authorization` scheme like `Token`.

We can have:
```
`mp.jwt.token.header`
`mp.jwt.token.cookie`
`mp.jwt.token.scheme`
```

vp642:
+1 to have a token in a cookie. I am working on a project that could have used this feature to make the code shorter.

radcortez:
If it is ok, I'll work on this and try to submit a PR.

radcortez:
I've pushed the spec update in #176. If we are happy with it, I'll write the required TCK tests.

sberyozkin:
#176 introduces support for the cookies and also makes it clear the other headers/schemes can be optionally supported. IMHO it would be better to resolve this issue now and instead open new issues going forrward dedicated to providing a 1st class support for the alternative well known schemes, headers (not just any headers). The spec is enough open enough for the implementers to support whatever they need and will back them up as needed.

PR #176, opened 2020-03-25
#93 - Added configurations to set JWT with a Cookie

radcortez:


spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -257,6 +257,25 @@ Reloading the Public Key from the location at runtime as well as the frequency o
...
+The `mp.jwt.token.scheme` config property allows to set up the HTTP Authentication Scheme which is expected to use with
+the Authorization header.
+
+When supplied, `mp.jwt.token.scheme` will override the standard `Bearer` scheme to retrieve the JWT token.
+
+#### `mp.jwt.token.cookie`
+
+The `mp.jwt.token.cookie` config property allows to set up the Cookie name which is expected to contain a JWT token.
+
+This config property must be used together with `mp.jwt.token.header` set to `Cookie` or else it will be ignored.
```
sberyozkin: @radcortez Perhaps `This property will be ignored unless 'mp.jwt.token.header' is set to 'Cookie'` ? Please also expand `config` -> `configuration` elsewhere.

sberyozkin:
Hi Roberto @radcortez Thanks for giving it a go, and good to see Github Actions in action :-).

I'm not sure right now about the custom schemes. May be we can add a requirement to support just approved `DPoP` say for 2.0 or later but in meantime, at the MP JWT level. we can start with suporting the users who need set JWT as cookies (which with the encryption on the way) will be not bad.

Also I'm thinking to restrict at the moment that `mp.jwt.token.header` can only have `Authorization` or `Cookie` to address the typical cases. That can be opened further in 2.0 or later. 

Sadly :-),  TCKs will have to be added. With the proposed restrictions, we can only have 1 more TCK added, select any test from the TCK jaxrs subpackage and have the token going with the cookie.

Thanks

spec/src/main/asciidoc/configuration.asciidoc
radcortez: Ok!

radcortez:
So, should just add `mp.jwt.token.header` and `mp.jwt.token.cookie` for now?

I was thinking that with `mp.jwt.token.header` you could also send whatever you want and read the JWT from there, to cover weird cases where the Authorization header might be coming in a different header. If we limit just to `Authorization` and `Cookie`, I think it doesn't cover the original issue in #93 .

What restrictions exactly? Can't we just add what we need?

sberyozkin:
@radcortez `Cookie` is the main alternative (at least it is obvious how the token may get into it).
If it is some random header then it is not clear what the expectations that we can set at the MP JWT level around these headers - does it represent some custom scheme too, etc. I know we do it in Smallrye, but to be honest, when I was adding a token.header option I was thinking about `Cookie` only.

The same about the scheme. If someone does `Authorization DPoP token` then we can't just say in the spec, get the token with the help of `mp.jwt.token.scheme` and process as usual as if it were `Bearer`, because DPoP involves the specific verification steps. So I'd rather have well-known schemes formally supported at the spec level in due time.

So the restrictions would be: `mp.jwt.token.header` can be set to `Cookie`, default is `Authorization`. And no `mp.jwt.token.scheme` is not supported.

We can open it easily for 2.0/etc once we have some concrete demand for more. For example a user would explain this is how I send a token, we can see if it fits into the spec constraints, etc. I hate to sound like a spec conservative :-), but unfortunately we can't go too fast in the spec :-)

radcortez:
I'm ok with this. My doubt here was the original issue resolution.

We are not resolving anything on the original issue, but adding a new feature to support JWT in a Cookie :)

sberyozkin:
@radcortez With `mp.jwt.token.header` we set up the spec framework, starting with Cookie, so it will be very much relevant to #93 (example, the user there expects a Cookie). Some other applications which use `Token` as a scheme can update to `Bearer` because this is exactly what a custom `Token` scheme actually represents :-).

radcortez:
Sure, I think this is still relevant.

My understanding of the original issue was that @jeanouii was suggesting to have it fully configurable and not tied to the existent schemes. So if your application would send the JWT in the header `foo` with Scheme `bar`, you could configure the application to read those instead. 

Of course, if you use an existent scheme like `Basic` this wouldn't work (or it shouldn't work). 

Anyway, I'm fine with the discussed approach. Let's add the cookie option for now.

radcortez:
@sberyozkin Check if this is better :)

spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -257,6 +257,22 @@ Reloading the Public Key from the location at runtime as well as the frequency o
...
 considered vendor-specific.
 
+#### `mp.jwt.token.header`
+
+The `mp.jwt.token.header` configuration property allows to set up the header which is expected to contain a JWT token.
+
+This configuration can either be `Authorization` (default), or `Cookie`. If a different configuration value is supplied,
+the configuration will be ignored and fallback to the default `Authorization`.
+
+Additional configuration options might be available in future versions of MP-JWT.
```
sberyozkin: +1 to this line, though I'd consider at the moment only refer to `mp.jwt.token.header`: `Other headers may be supported in future versions of MP-JWT` since it is not clear what other configuration options would be about.

spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -257,6 +257,22 @@ Reloading the Public Key from the location at runtime as well as the frequency o
 reloading is beyond the scope of this specification and any such feature should be
 considered vendor-specific.
 
+#### `mp.jwt.token.header`
+
+The `mp.jwt.token.header` configuration property allows to set up the header which is expected to contain a JWT token.
+
+This configuration can either be `Authorization` (default), or `Cookie`. If a different configuration value is supplied,
+the configuration will be ignored and fallback to the default `Authorization`.
```
sberyozkin: I believe it should fail with `DeploymentException` as otherwise MP JWT will be checking not what the application expects

spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -257,6 +257,22 @@ Reloading the Public Key from the location at runtime as well as the frequency o
...
+
+Additional configuration options might be available in future versions of MP-JWT.
+
+#### `mp.jwt.token.cookie`
+
+The `mp.jwt.token.cookie` configuration property allows to set up the Cookie name which is expected to contain a JWT
+token.
+
+This configuration must be used together with `mp.jwt.token.header` set to `Cookie` or else it will be ignored.
```
sberyozkin: Please update as agreed at https://github.com/eclipse/microprofile-jwt-auth/pull/176#discussion_r397833042 :-)

spec/src/main/asciidoc/configuration.asciidoc
sberyozkin: @radcortez Oh may be `Other headers and authentication schemes may be supported in future versions of MP-JWT` if you wanted to imply the schemes ?

spec/src/main/asciidoc/configuration.asciidoc
sberyozkin: Or if you prefer to keep it as is - then I'll personally be fine :-)

spec/src/main/asciidoc/configuration.asciidoc
radcortez: Sure.

spec/src/main/asciidoc/configuration.asciidoc
radcortez: Ops, sorry. Missed that part. Done.

sberyozkin:
Hey Roberto, @radcortez, we are getting there :-). It looks like this fragment is inserted where the location property is still described. Please reinsert it immediately after `mp.jwt.verify.issuer`.
Also, `Additional configuration headers and authentication schemes` might need to be changed. May be to just  `Other headers as well as alternative authentication schemes...` Sorry !

radcortez:
Ok, let see now :)

sberyozkin:
Hi Roberto @radcortez Adding some update to the spec is like peeling the onion :-), I keep discovering new things all the time with my PRs :-). `Names` class should also be updated.

The latest update is good, only one minor comment there, since it is a dedicated section which will affect the pending JWE PR too, it needs to be outside of `==`section dealing with the verification token properties, so `== JWT and HTTP headers` (and `===` for the properties) :-) and  I can tweak it further if needed.
Thanks

radcortez:
Oh yeah, there is more stuff missing that I plan to send in additional commits (API changes and TCK included). I just wanted for us to have an agreement in the actual spec before moving the changes further into the code.

Yes, I thought about the section titles as well. The issue is that we have a `### Configuration Parameters` title and I think everything that provided configuration properties should be under that umbrella.

What do you think?

sberyozkin:
@radcortez That `Configuration Parameters` section describes the properties related to token verification and is part of the signed tokens section, I'm mirroring it for the [decryption section](https://github.com/eclipse/microprofile-jwt-auth/pull/160/files#diff-b0600950ee868a147d4c30ea62a99094).
This PR introduces the properties which are orthogonal to how the token is secured, they are to do with the token delivery over HTTP. Hence they should be in a standalone section, as a sibling of the signed tokens section (and later - of the decryption one)
```
== JWT and HTTP Headers
=== Configuration properties
==== mp.jwt.token.header
```
would be better may be to have the same structure as the signed token one. We can also provide a separate doc ref listing all the config properties, but we don't have to do it with this PR

There is also a section to do with mapping the env properties but I'll reorganize it later.

Thanks

sberyozkin:
May be get that mapping properties too 
```
#### Mapping Properties to Environment Variables

Please see <<mapping-verification-props, Mapping Verification Properties to Environment Variables>> for the rules which also apply to the token header properties.

The portable environment variables names for the current MP-JWT token header properties are:

`mp.jwt.token.header` :: ...
`mp.jwt.cookie.name` :: ...
```

radcortez:
Ok, lets see now :)

I've already consolidated the mapping section to save you some work :)

sberyozkin:
@radcortez Nice, this looks good IMHO. 
Can you please to add a test suite to tck `jaxrs/*` ? One test would send a cookie jwt and may be another one in a cookie which does not match a configuration property

sberyozkin:
@rdebusscher Rudy, others, comments are welcome

radcortez:
Yes. Since we now have an agreement on the spec, I can reflect that on the rest of the project :)

sberyozkin:
@radcortez Hi Roberto, sorry, 
I've been contemplating and I think I proposed a bit too rigid response.
If a given implementor's customer says, look, we have `X-Authorization: token`, and we still want to use `mp.jwt.token.header` instead of your project specific property (ex, `smallrye.jwt.token.header`) then the implementors should have a spec's blessing for it. I still think we don't require supporting the other headers (or schemes) supported at this stage but make it optional. Spec already does it for the verification key sizes, so we need to keep the spec open enough and formalize as needed later on.

So here is a suggested update.
Instead of:
```
This configuration can either be `Authorization` (default), or `Cookie`. If a different configuration value is supplied, MicroProfile JWT implementations are required to throw a `DeploymentException`
Other headers as well as alternative authentication schemes may be supported in future versions of MP-JWT.
```
we simply say:
```
MP JWT implementations are required to support `Authorization` (default) or `Cookie` configuration values. 

Support for other headers or alternative authentication schemes is optional.  
```

This last sentence can stay unchanged across many versions of the spec, we will just increment by adding a `DPoP` support or some new header, etc... 

The other point, about `mp.jwt.cookie.name`, please add a default value:
```
The `mp.jwt.token.cookie` configuration property allows to set up the Cookie name (default is `Bearer`) which is expected to contain a JWT token.
```

I know Scott @starksm64 proposed `MP-JWT` earlier, David @dblevins reverted it, I'm not sure what was the context there :-), but `Bearer` as a default value reflects better the nature of such flows involving the cookies and aligns with a standard `Bearer` authorization scheme.

I know there is some bad press about JWTs being used as cookies - I need to review it and we then should offer some guidance. Most likely someone put the secrets into a signed JWT token cookie which could be easily introspected, which was then portrayed all over the web as a JWT flaw. We can likely add a recommendation to encrypt etc

Thanks

radcortez:
@sberyozkin No need to be sorry. This is a natural process :)

I was actually implementing the TCK right now and also came across with the possibility of setting a default name for the Cookie.

Also, I've came up with an ideia about relocating config names and fallbacks, so if a configuration name changes you can fallback to another name and make the life easier for implementations when you have these cases.

Anyway, I'm +1 on the proposed changes. Let me finish up here and I'll send an update.

sberyozkin:
@radcortez 

> Also, I've came up with an ideia about relocating config names and fallbacks, so if a configuration name changes you can fallback to another name and make the life easier for implementations when you have these cases.

What do you mean ? Configuration names are not going to change.

radcortez:
No, but for instance in SmallRye JWT we will need to support both `mp.jwt.token.header` and `smallrye.jwt.token.header`, right?

sberyozkin:
@radcortez This is not a spec problem

radcortez:
Yes, I know :)

But since you mentioned the spec vs implementation property name in your previous comment, I mentioned that :)

radcortez:
Anyway it gave me a good idea to propose to MP Config :)

sberyozkin:
@radcortez Right, I did mention a project specific property to make it clearer to the wider audience only 

> Anyway it gave me a good idea to propose to MP Config :)

This is good, perhaps the fallback idea an be generalized :-).

radcortez:
@sberyozkin 

Ok, in these last commit you will find:
- Updated spec text as we discussed in the last few comments.
- Updated Names.java
- TCK tests for the new configuration.

A few other notes:
- Do you think we should introduce a enum or some other structure to track the Authorization header possible values? At least the supported ones?
- We need to think in a way to make the TCK more config friendly. Right now, if you want to test a different configuration you require to create a new class and do the entire setup again.

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/CookieTokenTest.java
```diff
@@ -0,0 +1,167 @@
...
+        Assert.assertEquals(response.getStatus(), HttpURLConnection.HTTP_UNAUTHORIZED);
+        String reply = response.readEntity(String.class);
+        System.out.printf("Reply: %s\n", reply);
+    }
+
+
+    @RunAsClient
+    @Test(groups = TEST_GROUP_JAXRS,
+          description = "Validate a request with expired token in a Cookie fails with HTTP_UNAUTHORIZED")
+    public void expiredCookie() throws Exception {
```
sberyozkin: Having this test does not harm, but the token validaton (exp, woron issuer, etc) is already well covered, we are just testing here that the alt delivery channels work; if you'd like to keep it it is fine :-)

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/CookieTokenTest.java
```diff
@@ -0,0 +1,167 @@
...
+        URL publicKey = InvalidTokenTest.class.getResource("/publicKey.pem");
+        WebArchive webArchive = ShrinkWrap
+            .create(WebArchive.class, "CookieTokenTest.war")
+            .addAsManifestResource(new StringAsset(MpJwtTestVersion.MPJWT_V_1_2.name()), MpJwtTestVersion.MANIFEST_NAME)
+            .addAsResource(publicKey, "/publicKey.pem")
+            .addClass(TCKApplication.class)
+            .addClass(RolesEndpoint.class)
+            .addAsWebInfResource("beans.xml", "beans.xml")
+            .addAsManifestResource(config, "microprofile-config.properties");
+        System.out.printf("WebArchive: %s\n", webArchive.toString(true));
```
sberyozkin: I'm removing println in the new tests, I'll deal with the cleanup in the other tests later

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/TokenAsCookieIgnoredTest.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright (c) 2016-2020 Contributors to the Eclipse Foundation
```
sberyozkin: Only `2020` here in all the other test

sberyozkin:
Hi Roberto @radcortez it looks very neat, few minor comments only.

> Do you think we should introduce a enum or some other structure to track the Authorization header possible values? At least the supported ones?

Not sure, may be once we support `DPoP`

> We need to think in a way to make the TCK more config friendly. Right now, if you want to test a different configuration you require to create a new class and do the entire setup again.

Yeah, new TCK setup per every variation right now. If it can be optimized such that it works in various containers then it would be nice (recall for ex, multiple deployments do not work everywhere :-))

Thanks, and enjoy the weekend

sberyozkin:
@radcortez And one more test is needed, please update the existing `RolesAllowedTest` (which uses `Authorization`) that sending JWT with the default cookie name fails with 401.

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/CookieTokenTest.java
```diff
@@ -0,0 +1,167 @@
...
+    public void emptyCookie() throws Exception {
+        String token = TokenUtils.generateTokenString("/Token1.json");
+
+        String uri = baseURL.toExternalForm() + "endp/echo";
+        WebTarget echoEndpointTarget = ClientBuilder.newClient()
+                                                    .target(uri)
+                                                    .queryParam("input", "hello");
+        Response response = echoEndpointTarget
+            .request(TEXT_PLAIN)
+            .cookie("jwt", "")
```
sberyozkin: That may actually return 401

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/CookieTokenTest.java
radcortez: Yes, I know it was just to have a different test :)

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/CookieTokenTest.java
radcortez: Yeah, I don't like them either, but I've added them to keep consistency. If the idea is to remove them in the other tests, I can remove these right away :)

radcortez:
Hi @sberyozkin.

I believe I've fixed all the comments :)

radcortez:
@Emily-Jiang we had that discussion. The plan is to support Cookie based JWT for now, but implementations are free to support their own mechanisms.

sberyozkin:
@Emily-Jiang Yes, for 1.2 the idea is to keep the extra schemes/headers support optional. Once [DPoP](https://tools.ietf.org/html/draft-fett-oauth-dpop-04) is out - support it formally at the spec, etc. 
Providing a spec level support to do `DPoP` with a property alone won't do as we'd need to specify then how the verification works, etc :-). As for the other headers - the specification lets to support them but does not require, otherwise we'd also have to introduce some formality around it...Lets review it in the later versions, thanks

spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -357,8 +357,26 @@ return Public Key text in one of the supported formats.
...
+
+Support for other headers or alternative authentication schemes is optional.
+
+#### `mp.jwt.token.cookie`
+
+The `mp.jwt.token.cookie` configuration property allows to set up the Cookie name (default is `Bearer`) which is
+expected to contain a JWT token.
+
+This configuration will be ignored unless `mp.jwt.token.header` is set to `Cookie`.
+
```
sberyozkin: Hi @radcortez Referring to earlier point about providing some basic guidance, this is kind of out of scope how to create the secure cookies, but something simple like the following can be added here:
```
Providing the recommendations on how to securely use JWT tokens as a cookie is out of scope for the MP JWT specification but generally one should avoid putting a sensitive user information into a signed JWT token.
```

I think it is worth adding something like this to avoid some users saying MP JWT is careless in supporting the cookies. The issue of dealing with the cookies securely is not really a JWT related concern but providing a basic guideline would be fine I guess...

radcortez: Added!

sberyozkin:
@radcortez @rdebusscher thanks

PR #178, opened 2020-03-31
Fixes #120 - Inject empty JsonWebToken.

### LINKED ISSUE #120: Is it possible to @Inject Optional<JsonWebToken> ?
It would be really nice, if possible!

radcortez:


tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/EmptyTokenEndpoint.java
```diff
@@ -0,0 +1,44 @@
...
+public class EmptyTokenEndpoint {
+    @Inject
+    private JsonWebToken jsonWebToken;
+
+    @GET
+    @Path("/verifyEmptyToken")
+    public JsonObject verifyEmptyToken() {
+        boolean pass = jsonWebToken.getName() == null &&
+                       jsonWebToken.getClaimNames() == null &&
+                       jsonWebToken.getClaim("") == null;
```
sberyozkin, line 40: @radcortez please add `getRawToken` check

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -419,6 +419,12 @@ public class RolesEndpoint {
 
 ```
 
+If there is no JWT in the request, an empty `JsonWebToken` is injected.  Note that MP JWT will still perform
+Authentication and Authorization if the endpoint requires these verifications. Effectively, the injected empty token
+is only visible on unauthenticated and unauthorized endpoints.
```
sberyozkin, line 424: +1 - this is very precise

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/EmptyTokenEndpoint.java
radcortez: Done :)

spec/src/main/asciidoc/interoperability.asciidoc
sberyozkin: @radcortez Perhaps we need to add something here which would clarify why we even try to support it. For example: 
```
Particularly, the endpoints which need to control the authentication process themselves or only need to propagate the tokens further can check if a token is available by calling a `JsonWebToken.getRawToken()` method.
```
Something like that I guess.

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -419,6 +419,12 @@ public class RolesEndpoint {
...
+
+All method calls to the empty token return `null`.
```
Emily-Jiang, line 426: I suggest to move this next to the empty 'JsonWebToken'. 
an empty `JsonWebToken` is injected. =>
an empty `JsonWebToken` is injected, which means all method calls to this token return `null`.

spec/src/main/asciidoc/interoperability.asciidoc
sberyozkin: @radcortez I believe we need to add, immediately after Effectively, the injected empty token is only visible on unauthenticated and unauthorized endpoints. that
Token verification must still be performed if the token is available as I don't think we want to allow supporting the cases where anyone can drop whatever the want into JsonWebToken :-), of course we can't control anything at the MP JWT level spec, but if the users just want a raw token they they can just get Authorization header, but whatever goes into JsonWebToken must be verified.
The issue of not even validating the tokens for the public endpoints is really container specific. We can relax this requirement if there will be some pushback, but in 1.2 I believe we should start from a more conservative approach

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/EmptyTokenTest.java
```diff
@@ -0,0 +1,73 @@
...
+    @ArquillianResource
+    private URL baseURL;
+
+    @Deployment
+    public static WebArchive createDeployment() {
+        return ShrinkWrap
+            .create(WebArchive.class, "EmptyTokenTest.war")
+            .addAsManifestResource(new StringAsset(MpJwtTestVersion.MPJWT_V_1_2.name()), MpJwtTestVersion.MANIFEST_NAME)
+            .addClass(TCKApplication.class)
+            .addClass(EmptyTokenEndpoint.class)
```
sberyozkin, line 56: @radcortez As per the earlier comment IMHO we need 1 more test where 401 is returned if a valid token is provided

sberyozkin: Let me delete it for now till we agree on the text

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/EmptyTokenTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
+import static org.eclipse.microprofile.jwt.tck.TCKConstants.TEST_GROUP_JAXRS;
+
+public class EmptyTokenTest extends Arquillian {
+    @ArquillianResource
+    private URL baseURL;
+
+    @Deployment
+    public static WebArchive createDeployment() {
```
sberyozkin, line 51: Putting the suggestion back :-), so I believe we need 1 more test here, which would use the empty token endpoint (no RBAC annotations) but the archive also has the mp jwt properties, unless such TCK already exists where the token is injected into an endpoint without RBAC.

Please also update the base tck suite

radcortez:
Ok, added a few more tests and rewrote the spec part a bit.

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -419,6 +419,16 @@ public class RolesEndpoint {
 
 ```
 
+If there is no JWT in the request, an empty `JsonWebToken` is injected, which means all method calls to this token
+return `null`. Note that MP JWT will still perform Authentication and Authorization if the endpoint requires these
+verifications. Effectively, the injected empty token  is only visible on unauthenticated and unauthorized endpoints.
+
+If a JWT is sent to an endpoint that does not require Authentication and/or Authorization, the `JsonWebToken`
```
sberyozkin, line 426: @radcortez It looks good but I'd say it needs to be tweaked a bit, 
`If a JWT is sent to an endpoint that does not require Authentication and/or Authorization, its verification must still be performed before it is injected as `JsonWebToken`` or similar, given that `JsonWebToken` term is associated with an MP JWT interface

radcortez: Changed to `If a JWT is sent to an endpoint that does not require Authentication and/or Authorization, the `JsonWebToken`
verification must still be performed before being injected or retrieved from the `BeanManager`.`

sberyozkin: @radcortez IMHO there is no need to refer to Bean manager. Your original sentence looked good and simple, I've only asked to clarify that what should be validated is the token, `JsonWebToken` is an interface. We don't say for example `if a JsonWebToken is sent`...

radcortez: Ok, let me remove the last bit.

sberyozkin: @radcortez I'm sorry, but we start circling - may be it is my fault :-). All I'm asking for is that we don't say `JsonWebToken verification is still performed` because `JsonWebToken` is an interface. The implementations verify the raw tokens, not the interfaces, before making it accessible via `JsonWebToken`

Can we just say:
```
JWT must still be verified if it is sent to an endpoint that does not require Authentication and/or Authorization but has a `JsonWebToken` injection point.
```
?

radcortez: Sure, but that is why I've added the `BeanManager` portion in the first place. If I try ti retrieve and instance of `JsonWebToken` with the BeanManager, shouldn't that also be verified? The sentence seems to imply it only happens for injection points.

sberyozkin: @radcortez the phrase `JsonWebToken verification is still performed` does not make sense IMHO. Because `JsonWebToken` is a reserved word for the interface. So we don't verify the interfaces. And referring to some low level details like bean manager is also not needed. All I'm asking for is for a minor update to your original sentence, which will make it obvious that it is a token which still needs to be verified. Thanks

sberyozkin: @radcortez would it work for you if you just replace `JsonWebToken verification is still performed` with `JWT verification is still performed` given that use refer to a token as `JWT` in the same sentence ?

sberyozkin: No, that does not work - because it opens a new requirement that it is verified even if no injection point exists.
So perhaps the very 1st suggestion I made may work, whatever that can make it obvious without getting into the techicalities will do

radcortez: Yes, that is the issue that we need to figure out.

Does a token still gets validated if there is no injection point for `JsonWebToken`? What if I call CDI programatic API to retrieve the `JsonWebToken`. The one returned should also be valid.

That is why I believe that if we sent a token to an unprotected endpoint, it must always be verified. So there are two outcomes:

Unprotected resources
- IsTokenPresent -> validate, can warn in logs, but not fail the request, proceed to the endpoint
- TokenIsNotPresent -> proceed to the endpoint 

What do you think?

sberyozkin: @radcortez 

> Does a token still gets validated if there is no injection point for JsonWebToken

IMHO this is definitely a standalone separate issue. Lets complete this issue first. Thanks

radcortez: Ok, if we want to handle this is a separate issue, than I propose to drop the (or possible changes being discussed in the sentence):

> If a JWT is sent to an endpoint that does not require Authentication and/or Authorization, the `JsonWebToken` verification must still be performed.

And discuss and clarify the behaviour for the raised concern.

sberyozkin: @radcortez I disagree. The original issue has been about a case where a user has `JsonWebToken` injection point and would like to use this injection point. Therefore an issue about what to do when such injection point does not even exist is completely orthogonal. I'd like to have a sentence that makes it clear that when a `JsonWebToken` injection point exists the token if available must be verified.

radcortez: Yes, but we cannot just make the decision about validating if the injection point exists or not. I can go can call the CDI API programatically to retrieve an instance of the `JsonWebToken`. Shouldn't that be validated as well even if there is no injection point?

sberyozkin: @radcortez I'm not sure why are we are getting these lower-level technicalities, but either way, if the service code attempts to access a token via `JsonWebToken` `interface`, in whatever way they prefer, the verification of the token must happen. The minor update I suggested originally,
```
If a JWT is sent to an endpoint that does not require Authentication and/or Authorization then its verification must still be performed before it is injected as JsonWebToken``
```
had made it clear. If you think that CDI.getInstance  or whatever is not covered by this then

```
If a JWT is sent to an endpoint that does not require Authentication and/or Authorization then it still must be verified before it can be accessed via JsonWebToken interface`
```

certainly does IMHO

radcortez: Sorry, I guess we got a little bit lost through the discussion (at least me :) )

Yes. The second sentence should clear that concern. Thank you.

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -419,6 +419,16 @@ public class RolesEndpoint {
 
 ```
 
+If there is no JWT in the request, an empty `JsonWebToken` is injected, which means all method calls to this token
+return `null`. Note that MP JWT will still perform Authentication and Authorization if the endpoint requires these
+verifications. Effectively, the injected empty token  is only visible on unauthenticated and unauthorized endpoints.
+
+If a JWT is sent to an endpoint that does not require Authentication and/or Authorization then it still must be
+verified before it can be accessed via JsonWebToken interface`
```
sberyozkin, line 427: @radcortez please replace ``` with `.` at the end. And wrap `JsonWebToken` instead.
By the way, if you meant `CDI.getInstance` when asking what happens when no injection point is available then it was a very valid question - to me it is all the same I guess, `CDI.getInstance` implies getting in touch with the producer I guess, but I think it was useful to touch on it, and I hope the text leaves no ambiguity now, thanks

radcortez: Ops sorry. Wrong copy paste :)

No worries, is all semantics. The issue is that someone might interpret that validation would only occur if you have an injection point.

Issue #94, opened 2018-05-30
Introduce mapping for claims


jeanouii:
After creating a MP-JWT implementation in TomEE, I have done a small demo application using it so I could go and speed at conferences.

I found 2 things quite painful
- the 'aud' which is required and I'm not quite sure why
- the 'sub' which is also required.

Though I would be interesting in having details about why the 'aud' is required, I'd like to discuss here about the 'sub'.

The 'sub' claim is used to build the Subject so the security context can be used for all sort of things.
I understand the fact we need a name to build the subject but I don't get why we have introduced a new claim.

Most providers have already a claim with a name or something that can be used, so why adding another one?

I was wondering if now that we are working on configuration we could provide mapping for 
- principal claim, aka the name which would be 'sub' for background compatibility reasons
- roles/groups claim - choose the name of the claim used to build up the list of role principals (it could be groups by default, but why not roles or scopes?)

starksm64:
The potential difficulty will be with the scope of the mapping. Is it global, or does it need to be keyed off of some other claim like iss, or even a hash of the signer public key. Meaning, are there multiple MP-JWT issuers being seen in a given runtime implementation, or can the runtime microservices be reliably partitioned based on the MP-JWT issuer.

starksm64:
Is there a Map<String, String> capability in MP Config? That would make this relatively easy to do. Otherwise, I guess we could just use a property prefix that allowed us to build this up from a set of properties.

starksm64:
Let's get some usecases added so that we can start some TCK tests.

sberyozkin:
@jeanouii `aud` is always good to check because without it there is no way for the service to check this token is indeed intended for this service. Which among other things opens an attack space where the bad intermediary forwards the token intended originally/only for it to the next target, etc.
We are bringing an option to let the users enforce the audience if they need to.

We can keep this issue open though to review how the Mappers can be introduced to map the role information when no `groups` is available

PR #179, opened 2020-04-03
Make groups claim optional

### LINKED ISSUE #129: Requiring the 'groups' claims is too restrictive
I have just seen a user reporting that the MP-JWT runtime fails to validate Auth0 tokens. I think it is a major barrier to the wider adoption of MP JWT given that many popular OIDC providers will not be able to include the `groups`claim.
IMHO the 'group' claim should be treated somehow similarly to the (required) `upn` claim. For example, one option could be for the MP JWT implementation to default to 'Default' groups or some other default value.
Rejecting the otherwise perfectly valid tokens used in the productions to secure the services is too restrictive

sberyozkin:
This PR updates `interoperability` document and adds a simple test (`PermitAll` is really equivalent to requiring no RBAC).

Fixes #129.

CC @radcortez, Roberto, not sure why I can't request a review from you, probably need to update some permissions

### DIFF
--- a/spec/src/main/asciidoc/interoperability.asciidoc
+++ b/spec/src/main/asciidoc/interoperability.asciidoc
@@ -33,10 +33,10 @@ To meet those requirements, we introduce 2 new claims to the MP-JWT:
 
 * "upn": A human readable claim that uniquely identifies the subject or user principal of the token, across
 the MicroProfile services the token will be accessed with.
-* "groups": The token subject's group memberships that will be mapped to Java EE style application
-level roles in the MicroProfile service container.
+* "groups": A list of group names to be assigned to the principal of the MP-JWT. This typically will require a mapping at the application container level to application roles. A one-to-one mapping between group names and application role names is done by default when no custom mapping is defined in an implementation specific way.
 
-### Minimum MP-JWT Required Headers and Claims
+### Minimum MP-JWT Header and Claim Requirements
+
 The required minimum set of MP-JWT headers and claims is:
 
 typ:: This JOSE header parameter identifies the token as an RFC7519 and must be "JWT" https://tools.ietf.org/html/rfc7519#section-5.1[RFC7519, Section 5.1]
@@ -61,11 +61,10 @@ upn:: This MP-JWT custom claim is the user principal name in the java.security.P
     principal name in javax.security.enterprise.identitystore.IdentityStore. If this claim is missing, fallback to
     the http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims["preferred_username", OIDC Section 5.1] should be
     attempted, and if that claim is missing, fallback to the "sub" claim should be used.
-groups:: This MP-JWT custom claim is the list of group names that have been assigned to the principal of the MP-JWT.
-    This typically will require a mapping at the application container level to application deployment roles, but a
-    one-to-one between group names and application role names is required to be performed in addition to any
-    other mapping.
 
+[NOTE] It is recommended that JWT tokens have a `groups` claim if the endpoint requires Authorization but MP JWT implementations can map the groups from the other claims if the tokens have been issued by OpenId Connect and other providers which currently do not support MP JWT.
+
+If no groups information can be extracted directly from the `groups` claim or via the custom mappers from other custom claims in a given token then this token can be accepted if the endpoint requires Authentication only.
 
 [NOTE]
 NumericDate used by `exp`, `iat`, and other date related claims is a JSON numeric value
@@ -219,6 +218,17 @@ public interface JsonWebToken extends Principal {
     default <T> Optional<T> claim(String claimName) {
         return Optional.ofNullable(getClaim(claimName));
     }
+
+    /**
+     * A utility method to access a claim value in an {@linkplain Optional}
+     * wrapper
+     * @param claim - the claim
+     * @param <T> - the type of the claim value to return
+     * @return an Optional wrapper of the claim value
+     */
+    default <T> Optional<T> claim(Claims claim) {
+        return claim(claim.name());
+    }
 }
 ```
 
@@ -334,6 +344,8 @@ public enum Claims {
 }
 ```
 
+Note that the `groups` and `aud` claims should only be injected into a Set of String, not as a comma delimited string.
+
 The current complete set of valid claim types is therefore, (excluding the invalid Claims.UNKNOWN Void type):
 
 * java.lang.String
@@ -384,10 +396,10 @@ invalid MP-JWT token if any of the following conditions are not met:
 
 1. The JWT must have a JOSE header that indicates the token was signed using the RS256 algorithm when the service endpoint expects signed tokens.
 2. The JWT must have a JOSE header that indicates the token was encrypted using the RSA-OAEP and A256GCM algorithms when the service endpoint expects encrypted tokens. If the encrypted content is a nested signed JWT token then it must meet the condition 1.
-2. The JWT must have an iss claim representing the token issuer that maps to an MP-JWT implementation
+3. The JWT must have an iss claim representing the token issuer that maps to an MP-JWT implementation
 container runtime configured value. Any issuer other than those issuers that have been whitelisted
 by the container configuration must be rejected with an HTTP_UNAUTHENTICATED(401) error.
-3. The JWT signer must have a public key that maps to an MP-JWT implementation container runtime
+4. The JWT signer must have a public key that maps to an MP-JWT implementation container runtime
 configured value when service endpoint is configured to verify the signed tokens and the JWT encryptor must have a private key that maps to an MP-JWT implementation container runtime configured value when service endpoint is configured to decrypt the encrypted tokens. Any public or private key other than those that have been whitelisted by the container configuration must be rejected with an HTTP_UNAUTHENTICATED(401) error.
 
 [NOTE]
--- a/tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/RolesAllowedTest.java
+++ b/tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/RolesAllowedTest.java
@@ -89,7 +89,7 @@ public class RolesAllowedTest extends Arquillian {
 
     @BeforeClass(alwaysRun=true)
     public static void generateToken() throws Exception {
-        token = TokenUtils.generateTokenString("/Token1.json", null);
+        token = TokenUtils.generateTokenString("/Token1.json");
     }
 
     @RunAsClient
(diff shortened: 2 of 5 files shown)


sberyozkin:
@rdebusscher FYI, I've moved the text to do with the `aud` injection hint here, immediately under `Claims` description as part of this PR. Perhaps it will be better here. Also update `JsonWebToken` description

tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/RolesAllowedTest.java
```diff
@@ -143,6 +143,24 @@ public void callEcho() throws Exception {
         Assert.assertEquals(reply, "hello, user=jdoe@example.com");
     }
 
+    @RunAsClient
+    @Test(groups = TEST_GROUP_JAXRS,
+        description = "Validate a request with MP-JWT without a groups claim succeeds with HTTP_OK}")
+    public void callEchoNoGroups() throws Exception {
+        Reporter.log("callEcho, expect HTTP_OK");
```
sberyozkin, line 150: Oops, I need to fix it, sorry, I thought I did

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -33,11 +33,11 @@ To meet those requirements, we introduce 2 new claims to the MP-JWT:
 
 * "upn": A human readable claim that uniquely identifies the subject or user principal of the token, across
 the MicroProfile services the token will be accessed with.
-* "groups": The token subject's group memberships that will be mapped to Java EE style application
-level roles in the MicroProfile service container.
+* "groups": A list of group names that have been assigned to the principal of the MP-JWT. This typically will require a mapping at the application container level to application deployment roles, but a one-to-one between group names and application role names is required to be performed in addition to any other mapping when this claim is available in the token.
```
radcortez, line 36: Maybe change `that have been assigned` to `to be assigned`?

Change `application deployment roles` to just `application roles`.

radcortez:
No worries. I think I don't have permissions yet. I know that my committer status was approved, but not sure how long it takes for the process to finish.

spec/src/main/asciidoc/interoperability.asciidoc
sberyozkin: @radcortez This is the original spec text. I'm assuming `have been assigned` means that these groups have already been allocated to a given client/user somewhere in the OIDC admin console (ex Keycloak) and next - they may need to be mapped to the deployment specific roles.
 
`application deployment roles` -> `application roles` - fine with me

radcortez: I interpret that as the token containing the `groups` and being used to produce the `Principal`. Because `groups` already exist in the token before the `Principal` I think it would be more clear as a future action `to be assigned`.

I guess it is only semantics :) I'm fine either way.

sberyozkin: Sure, I've no problems really too, so will update

sberyozkin:
@rdebusscher, others, any comments on your end ? I'll be happy to go with Roberto's comments otherwise, even though I can't technically assign him as a reviewer yet.
Thanks

sberyozkin:
I'd like to follow up with the clarification around few other claims immediately after this PR

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -33,11 +33,11 @@ To meet those requirements, we introduce 2 new claims to the MP-JWT:
 
 * "upn": A human readable claim that uniquely identifies the subject or user principal of the token, across
 the MicroProfile services the token will be accessed with.
-* "groups": The token subject's group memberships that will be mapped to Java EE style application
-level roles in the MicroProfile service container.
+* "groups": A list of group names to be assigned to the principal of the MP-JWT. This typically will require a mapping at the application container level to application roles, but a one-to-one between group names and application role names is required to be performed in addition to any other mapping when this claim is available in the token.
```
rdebusscher, line 36: Last part can be confusing because we basically say "Some other mapping is **required** on the group names". but we do not specify what that means. But use the word **required**.

I guess we want to leave the option open that the runtime performs any custom mapping on the group names which are configured (implementation specific)

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -61,11 +61,8 @@ upn:: This MP-JWT custom claim is the user principal name in the java.security.P
     principal name in javax.security.enterprise.identitystore.IdentityStore. If this claim is missing, fallback to
     the http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims["preferred_username", OIDC Section 5.1] should be
     attempted, and if that claim is missing, fallback to the "sub" claim should be used.
-groups:: This MP-JWT custom claim is the list of group names that have been assigned to the principal of the MP-JWT.
-    This typically will required a mapping at the application container level to application deployment roles, but a
-    one-to-one between group names and application role names is required to be performed in addition to any
-    other mapping.
 
+[NOTE] It is recommended that JWT tokens contain a `groups` claim if the endpoint requires Authorization. Tokens without this claim can be accepted if the endpoint requires Authentication only. Support for mapping the groups from other claims is optional.
```
rdebusscher, line 65: Shouldn't we introduce the mapper SPI or the configuration that the groups can be taken from another claim here also.

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -381,12 +391,12 @@ The MP-JWT specification requires that an MP-JWT implementation reject a bearer
 invalid MP-JWT token if any of the following conditions are not met:
 
 1. The JWT must have a JOSE header that indicates the token was signed using the RS256 algorithm.
-2. The JWT must have an iss claim representing the token issuer that maps to an MP-JWT implementation
-container runtime configured value. Any issuer other than those issuers that have been whitelisted
-by the container configuration must be rejected with an HTTP_UNUATHENTICATED(401) error.
-3. The JWT signer must have a public key that maps to an MP-JWT implementation container runtime
+2. The JWT signer must have a public key that maps to an MP-JWT implementation container runtime
```
rdebusscher, line 394: `JWT signer` is not a good term as we are not signing a token here but verifying it.

JWT Signature verifier?

Or what exactly we do want to tell the people here?  When a key identification (kid then) is encountered which is not expected we need to return 401?

sberyozkin: @rdebusscher I've just moved the existing lines around there, the whole `interoperability.doc` needs a good refresh but `JWT Signature verifier` is minor but a good start. I believe `kid` is mentioned elsewhere in scope of the JWK format text, this is just a high level requirement that a public key should be available.

spec/src/main/asciidoc/interoperability.asciidoc
sberyozkin: @rdebusscher #94 was opened awhile back, but I've marked it now for JWT 2.0. I'd like to release `1.2` asap so that we can go ahead with breaking things and try to make it for MP 4.0 with MP JWT 2.0. This is why I'm postponing everything that can be postponed including the builder API, even though it does not break anything. The only other minor improvement I'd like to try to squeeze is the ES256 support in order to ensure that when the builder api arrives at some point it will be synced (as per your earlier comments) with the server side support.
So, it was a long way to suggest to address the mapper spi post 1.2 :-)

spec/src/main/asciidoc/interoperability.asciidoc
sberyozkin: @rdebusscher I've read that text and it just requires that a signer which signs (with a private key) should have a matching public key that can be mapped by MP JWT server somehow. We cam tweak for sure, but I'd rather have a dedicated issue (Improve/clarify the texts etc) as it can easily take few days :-)

sberyozkin:
@rdebusscher I've resolved 2 comments (please reopen if preferred) and kept one open - please check the updated text - I've updated it to make it obvious MP JWT implementations can provide the custom mappers and mentioned these mappers will be supported formally in the future versions

sberyozkin:
@rdebusscher I've lifted the change request here as well in preparation for a merge. As said in the other PR, I will hold off till Fri and we can always review/reopen before 1.2 is out

spec/src/main/asciidoc/interoperability.asciidoc
sberyozkin: @rdebusscher I've added a note below that the custom mapping is possible. The above just says that of no mapping is done then the groups values are taken as is.

rdebusscher: I still find it confusing (the one-to-one mapping AND any other mapping is required). But everyone else if fine with this, you can merge it.

Proposal 

>This typically will require a mapping at the application container level to application roles, a one-to-one between group names and application role names is the default when no custom mapping is defined in an implementation specific way.

sberyozkin:
@rdebusscher Sure, Let me tweak it and I'll ping you once it is ready

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -33,10 +33,10 @@ To meet those requirements, we introduce 2 new claims to the MP-JWT:
 
 * "upn": A human readable claim that uniquely identifies the subject or user principal of the token, across
 the MicroProfile services the token will be accessed with.
-* "groups": The token subject's group memberships that will be mapped to Java EE style application
-level roles in the MicroProfile service container.
+* "groups": A list of group names to be assigned to the principal of the MP-JWT. This typically will require a mapping at the application container level to application roles. A one-to-one mapping between group names and application role names is done by default when no custom mapping is defined in an implementation specific way.
```
sberyozkin, line 36: @rdebusscher How does it look ? It does look clearer than the original text, I agree :-) CC @radcortez

Emily-Jiang: We should say comma separated group names.

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -61,11 +61,10 @@ upn:: This MP-JWT custom claim is the user principal name in the java.security.P
     principal name in javax.security.enterprise.identitystore.IdentityStore. If this claim is missing, fallback to
     the http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims["preferred_username", OIDC Section 5.1] should be
     attempted, and if that claim is missing, fallback to the "sub" claim should be used.
-groups:: This MP-JWT custom claim is the list of group names that have been assigned to the principal of the MP-JWT.
-    This typically will require a mapping at the application container level to application deployment roles, but a
-    one-to-one between group names and application role names is required to be performed in addition to any
-    other mapping.
 
+[NOTE] It is recommended that JWT tokens have a `groups` claim if the endpoint requires Authorization but MP JWT implementations can map the groups from the other claims if the tokens have been issued by OpenId Connect and other providers which currently do not support MP JWT. Mapper SPI for extracting the groups from the custom claims will be supported in the future versions of this specification in order to make MP JWT implementations more interoperable.
```
Emily-Jiang, line 65: I suggest to delete "Mapper SPI for extracting the groups from the custom claims will be supported in the future versions of this specification in order to make MP JWT implementations more interoperable." as it muddles the water and no point to put some future features in the current spec as we don't have control which future version it will end up with or maybe never.

spec/src/main/asciidoc/interoperability.asciidoc
rdebusscher: looks goods to me now.

sberyozkin: @Emily-Jiang 

> We should say comma separated group names.

This will add to the confusion IMHO. `groups` claim is an array, so adding the `comma-separated` would not work well I believe. I'm not sure where commas would apply - it is an implementation specific aspect I guess. Thanks

spec/src/main/asciidoc/interoperability.asciidoc
sberyozkin: @Emily-Jiang sounds good

sberyozkin:
Going ahead with the merge now - any concerns - please open new issues. Thanks all

Issue #27, opened 2017-08-31
Both scopes or one scope?


johnament:
We had a similar issue on the JMS spec, and it leads to a potentially much more complex implementation

https://github.com/eclipse/microprofile-jwt-auth/blob/master/spec/src/main/asciidoc/interoperability.asciidoc#cdi-injection-requirements

In this section, its declared that we need to support injection of either RequestScoped or Dependent scoped identity.  Is dependent really appropriate here?  What is the use case?

I would prefer to see request scoped only, unless there's some specific need.

starksm64:
A producer for a javax.injection.Provider<String> requires Dependent scope due to the unproxyable nature of String. I raised an issue with the CDI spec on why the javax.injection.Provider is not essentially a wrapper, thus creating a proxyable type and did not get a satisfactory answer.

starksm64:
So based on today's MP-JWT call, what this issue comes down to is the following:
1. The JsonWebToken needs @RequestScoped scoping.
2. The associated claim values need @Dependent scoping and implementations should verify that injection of a raw claim value into a non-requestscoped context results in an error or warning.

johnament:
Sorry I had to head out early.  What I would recommend:

- Optional, String, Long, Date etc injection points have scope dependent
- All other injection points have request scope

It would be preferable that we don't even call out injection of provider & instance, since those are built in CDI constructs.

starksm64:
Should we just say any non-proxyable type as defined by the CDI spec must have dependent scope, and list a few example like you have mentioned?

johnament:
I guess its a question of what we consider to be the finite list of fields injectable with `@Claim`.  I think the spec should clarify that.

What I'm now seeing is that in order to create any `@Claim` injection point, I need the JWT and the requested claim, which means I always need dependent scope beans.

Emily-Jiang:
I agree this JsonWebToken bean should not have two scopes. However, I think it should match to the in-built Principal bean provided by CDI, because JsonWebToken is an extension to Principal. In CDI spec, there is a builtin Principal bean provided and it is Dependent scope. I ran this past Weld-dev team. Martin Kouba confirmed _"It's @Dependent. But a special proxy is injected and SPI 
org.jboss.weld.security.spi.SecurityServices.getPrincipal() is invoked 
to obtain the current caller identity for every invocation of an 
injected reference. Well, it works similar to how @Dependent stateless session beans work. 
[...]

PR #180, opened 2020-04-05
Add ApplicationScoped injection test

### LINKED ISSUE #45: `@ApplicationScoped` beans may use `@Claim ClaimValue` injection
We have no tests that show an `@ApplicationScoped` bean using `@Claim ClaimValue` injection.

This is supported and should be tested

sberyozkin:
Fixes #45.

I'm not really sure it can be done for 1.2 given `interoperability.adoc`:

```
#### Handling of Non-RequestScoped Injection of Claim Values
MP-JWT implementations are required to validate the use of claim value injection into contexts that do not match
@RequestScoped. When the target context has @ApplicationScoped or @SessionScoped scope, implementations are
required to generate a `javax.enterprise.inject.spi.DeploymentException`. For any other context, implementations
should issue a warning, that may be suppressed by an implementation specific configuration setting.

[NOTE]
If one needs to inject a claim value into a scope with a lifetime greater than @RequestScoped, such as
@ApplicationScoped or @SessionScoped, one can use the javax.enterprise.inject.Instance interface to do so.
```

Can CDI experts recommend what can be simplified there, throwing `DeploymentException` seems wrong. I think this should be dropped


rmannibucau:
+1, there is not a single technical reason in CDI or MP for that restriction to exist (can only be a wrong impl leak in the spec if justified by anything).

radcortez:
Wasn't this added to make injection of ClaimValue consistent with primitives?

rmannibucau:
@radcortez I always thought it was the opposite, ClaimValue wrapper was added to bypass primitive limitations, :thinking:

radcortez:
Yes, that is true.

But I think that was added because we do support injection of plain primitive / wrapper types, plus the ClaimValue injections. So, if you wrap your claim in a `ClaimValue` in an `ApplicationScoped` bean it would work, but not a primitive or wrapper type (with unexpected results). 

Most likely, we shouldn't have supported primitives / wrappers in the first place.

rmannibucau:
Guess the discussion is to pick something between these lines:

1. Keep it as it is (clearly i'm not fan).
2. Enable to inject in app scoped bean at the user risk (i'm fine with it, if he is using CDI he knows the implications and already has that with config so not a bad compromise).
3. Rethink the way we design injections and just drop ClaimValue, user will have to create a UserRequestScoped class with all injections with request scope scope (surely clean on an API point of view but highly wrong on a perf one).

Note that 3 has a variant of "just provide JsonWebToken instance injectable anywhere" which is more or less the same without the perf impact but a very light API not that user friendly.

radcortez:
> 2. Enable to inject in app scoped bean at the user risk (i'm fine with it, if he is using CDI he knows the implications and already has that with config so not a bad compromise).

Unfortunately, we cannot assume everybody knows the implications. I think we need to figure out a way so the user doesn't shoot himself in the foot. Config is not an issue, because configurations are usually not request scopes like the JWT, so it is less likely you are going to have issues over there.

Maybe we can consider two additional options:

1. Drop the requirement for `ClaimValue` and specify that the validation happens only for primitive / wrapper injection types. Instead of a `DeploymentException` log a warning.

2. Drop primitive / wrapper injection type completely.

sberyozkin:
What risks are involved when `ClaimValue` is injected into `ApplicationScoped` beans ? `ClaimValue` is an interface, so I'm not really following why `ClaimValue<String> rawToken` does not work, we had a test in Thorntail which was done exactly the way this PR's test is done, verifying that no stale claim values are picked up.

rmannibucau:
> Unfortunately, we cannot assume everybody knows the implications. 

Then you want to change it in mp-config too? Exact same issue happens there since ConfigSource are mutable by design.

I'm -1 to drop primitive support if we keep ClaimValue injection. Others are acceptable IMHO.

> What risks are involved when ClaimValue is injected into ApplicationScoped beans ?

None. Same in geronimo-jwt-auth, no technical issue with that.

radcortez:
> Then you want to change it in mp-config too? Exact same issue happens there since ConfigSource are mutable by design.

In config is less likely to happen, but I get your point.
 
> I'm -1 to drop primitive support if we keep ClaimValue injection. Others are acceptable IMHO.

I'm fine to drop `ClaimValue` support and keep only primitive types injection and use `Instance` if required. There are too many options on how to inject stuff that just complicates implementations and confuse users. 

> What risks are involved when ClaimValue is injected into ApplicationScoped beans ?

None, because `ClaimValue` is proxyable so it will always retrieve the right value for the injected claim in the request. I guess that was the reason why `ClaimValue` was added, to workaround the non proxyable primitive types.

radcortez:
The only reason I see for that to be done is because if you have:

```
@ApplicationScoped
class Resource {
  @Claim("claim")
  @Inject
  ClaimValue<String> claim // work as expected. proxy the call to the request
}
```

Change it to:

```java
@ApplicationScoped
class Resource {
  @Claim("claim")
  @Inject
  String claim // injection works, but most likely you get the value from the first token sent and never changes.
}
```

So I guess the requirement was added for the user experience to be the same.

radcortez:
A past discussion:
https://github.com/eclipse/microprofile-jwt-auth/issues/27

rmannibucau:
Side note: 

> I'm fine to drop ClaimValue support and keep only primitive types injection and use Instance if required.

This one is not that fine I fear. ClaimValue being provided by the extension can be made fast, Instance (and Provider) require by spec a resolution+instantiation so even if CDI container caches, it implies several steps which are slow/impacting for request scope instances.


What about:

1. Keep API like we have today,
2. Relax injections in application scoped beans (= enable as being not portable but not forbidden).

Can be enough as a first step.

radcortez:
> 2. Relax injections in application scoped beans (= enable as being not portable but not forbidden).

I would say to remove the restriction entirely for any bean type and optionally, implementations can issue a warning for non RequestScoped types if the injection point is a primitive or wrapper type.

sberyozkin:
@radcortez, @rmannibucau, thanks.

I'll just indeed remove the restriction and recommend that the implementations should warn the users when they have `String`/etc injected into `Application`, along the lines of of what @radcortez suggests. I guess making it optional would also work but encouraging to warn seems better.
The same issue exists in JAX-RS when a field injection of the pathparam/etc is done into singletons and none of the implementations  fails to start AFAIK but may warn.

I'll ping both of you once I'll open it for review

sberyozkin:
HI @rmannibucau, @radcortez, @rdebusscher, how does it look now ? It is open for review

Thanks

rmannibucau:
@sberyozkin two small details for me:

1. I wouldn't mention "should log" but "is encouraged to log". I know it is a detail but in spec "should" is the step before the "must" and there is no real way today we validate logging so probably not that bad to clarify it is really a will but can end up as no-op too,
2. If I didn't miss something you never call twice the application scoped endpoint to check the ClaimValue "moves" with different incoming jwt.

Otherwise it looks like it fixes this issue to me.

radcortez:
> 1. I wouldn't mention "should log" but "is encouraged to log". I know it is a detail but in spec "should" is the step before the "must" and there is no real way today we validate logging so probably not that bad to clarify it is really a will but can end up as no-op too,

Or just say "may optionally log" or simple "may log".

> 2. If I didn't miss something you never call twice the application scoped endpoint to check the ClaimValue "moves" with different incoming jwt.

There is a `verifyInjectedRawToken2Provider()` with a different token. Should this be enough?

rmannibucau:
@radcortez +1 for "may", about `verifyInjectedRawToken2Provider`, I was thinking about a method calling twice the endpoint and checking the token was actually different for both but not relying on 2 different methods.

sberyozkin:
@rmannibucau @radcortez Thanks.
Romain, I've updated the tests to have a single test making 2 calls, and I've also updated the endpoint to echo the injected token back which is checked on the client side.
How does it look now ?

sberyozkin:
@rmannibucau Cool, thanks, I prefer to have at least one non-Redhat implementer's feedback, so your approval in the last comment is sufficient. I'll wait in case @rdebusscher has something to add but will merge later tomorrow
Romain, Roberto, thanks

sberyozkin:
@rdebusscher, others Merging now, if someone has any concerns then reopen the linked issue/etc, we'll sort out the finer details if needed

PR #185, opened 2020-04-14
Fixed tests with new expected size of claimMap, due to the add of a boolean property.

radcortez:


sberyozkin:
@radcortez cool, thanks

PR #186, opened 2020-04-14
Clarify how a claim value injection is handled for SessionScoped beans

### LINKED ISSUE #183: Clarify ClaimValue in combination with @SessionScoped
Now that `ClaimValue` can be 'injected' into Application and Session scoped beans (See changes in spec doc in PR #180), the usage of SessionScoped beans can lead to exception as `ClaimValue` is not `Serializable`.

sberyozkin:
Fixes #183.

I've restored the bit related to DeploymentException and SessionScoped beans with some minor restructuring. Also dropped some old note, there is no much point to keep it just to keep promising the exp grace may be added, we can add it anyway. 
CC @radcortez 

Given the update I do think we'd be better off with just dropping a `SessionScoped` baggage, a warning is good enough. If it works for some other bigger scopes whatever they are (where NPE etc may also happen) then it works for SessionScoped. 

Thoughts ?

### DIFF
--- a/spec/src/main/asciidoc/interoperability.asciidoc
+++ b/spec/src/main/asciidoc/interoperability.asciidoc
@@ -402,11 +402,6 @@ by the container configuration must be rejected with an HTTP_UNAUTHENTICATED(401
 4. The JWT signer must have a public key that maps to an MP-JWT implementation container runtime
 configured value when service endpoint is configured to verify the signed tokens and the JWT encryptor must have a private key that maps to an MP-JWT implementation container runtime configured value when service endpoint is configured to decrypt the encrypted tokens. Any public or private key other than those that have been whitelisted by the container configuration must be rejected with an HTTP_UNAUTHENTICATED(401) error.
 
-[NOTE]
-A future MP-JWT specification may define how an MP-JWT implementation makes use of the MicroProfile Config
-specification to allow for configuration of the issuer, issuer public key and expiration clock
-skew. Additional requirements for validation of the required MP-JWT claims may also be defined.
-
 ## Mapping MP-JWT Tokens to Java EE Container APIs
 
 The requirements of how a JWT should be exposed via the various Java EE container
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -622,11 +617,17 @@ string or a `Claims` enum value. The string form would allow for specifying non-
 claims while the `Claims` enum approach guards against typos.
 
 #### Handling of Non-RequestScoped Injection of Claim Values
+
+MP-JWT implementations are required to support a claim value injection into an @ApplicationScoped scope. A warning may be issued when the @ApplicationScoped injection point is a primitive or wrapper type.
+
+MP-JWT implementations are required to generate a `javax.enterprise.inject.spi.DeploymentException` for a claim value injection into a @SessionScoped scope.
```
rdebusscher, line 623: We should use the term `Passivation capable beans, for example SessionScoped` to be entirely correct.

The sentence about `lifetime greater than @RequestScoped.` is not needed in that case.

radcortez: +1

sberyozkin: @rdebusscher Sure, so you mean to have something like
```
MP JWT implementations may issue a warning for Passivation capable beans, for example, SessionScoped context.
```
Then I'll drop a requirement to throw an exception if we allow for a warning in case of SessionScoped ? It would be my preference for sure

rdebusscher: not the warning, the exception when used on a Passivation capable bean.

sberyozkin: @rdebusscher Please check the updated text...

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -634,11 +629,17 @@ string or a `Claims` enum value. The string form would allow for specifying non-
 claims while the `Claims` enum approach guards against typos.
 
 #### Handling of Non-RequestScoped Injection of Claim Values
+
+MP-JWT implementations are required to support a claim value injection into an @ApplicationScoped scope. A warning may be issued when the @ApplicationScoped injection point is a primitive or wrapper type.
```
radcortez, line 633: I would probably rewrite this a little bit to:

MP-JWT implementations are required to support a claim value injection into  **@ApplicationScoped scoped beans**. A warning may be issued when the injection point is a **primitive type (or equivalent wrapper)**.

sberyozkin: @radcortez What about `String` ?

radcortez: In practice byte[] is a String and the primitive type :D

Is was just to clarify that the wrappers refer to the wrapper of the primitive type and not just any wrapper. The original text might give the impression that any wrapper type is supported.

sberyozkin: Perhaps it is simpler `A warning may be issued when the injection point is not a 'ClaimValue' or 'Provider' interface`.

radcortez: Sounds good.

sberyozkin:
@radcortez @rdebusscher How does it look now ?

radcortez:
Looks good!

sberyozkin:
@radcortez @rdebusscher thanks

PR #188, opened 2020-04-24
Support ES256 signatures

### LINKED ISSUE #161: Support ES256 signatures
Right now only `RS256` is supported. However it is known that some providers/intermediaries will use `ES256`, for example, `AWS Load Balancer` which is referenced in #smallrye/smallrye-jwt/issues/143, but also independently confirmed in #auth0/node-jsonwebtoken/issues/514 and possibly in others.

Even without those issues, supporting a well known alternative to `RS256` (FYI, it has a higher `Recommended+` status compared to `RS256` in the [JWA spec](https://tools.ietf.org/html/rfc7518#section-3.1) - which is only informative but indicates that the experts regard it quite highly) would let MP-JWT cover a wider user audience.

it will also sync it with the builder API PR.

My understanding is that no major changes will be needed to support both `RS256` and `ES256`. The PR will recommend using only `RS256` or `ES256` but not both at the same time (it is considered to be a risky idea in JOSE for the token `alg` header alone dictating indirectly what key to load), which will let the implementations decide, when loading the keys, if they should be treated as RS or EC keys without any ambiguities.

sberyozkin:
Fixes #161

This PR introduces a support for the ECDSA ES256 signatures.
The specification text updates are quite straightforward (note I also did a few minor fixes around the JWE related text). 
The only 'major' update is a new property `mp.jwt.verify.publickey.algorithm` - the justification in the PR text but in short it is consistent with the conservative approach taken when JWE was introduced: a given endpoint's security policy is not expected to be dynamic - it is unlikely the secure services will want to process all sort of signature algorithms in scope of a single endpoint. This new property can take `ALL`  (algorithms) in due time if there will be some real practical requirements.

I'll do the TCK tests next and open this PR for review


sberyozkin:
@rdebusscher @starksm64 This PR is now open for review.
I've added 3 TCK tests which is more than enough IMHO to start with.
Note that in `PublicKeyEndpoint` I've skipped checking that `ECPublicKey` can be created from the JWK key - it is quite hard to do without a utility code which I believe should not be used in the TCK endpoints, but I'm not sure it is even needed - the content of the EC JWK key is verified.

spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -238,14 +251,23 @@ MicroProfile JWT implementations are required to throw a `DeploymentException` i
...
 Reloading the Public Key from the location at runtime as well as the frequency of any such
 reloading is beyond the scope of this specification and any such feature should be
 considered vendor-specific.
 
+#### `mp.jwt.verify.publickey.algorithm`
+
+The `mp.jwt.verify.publickey.algorithm` configuration property allows for specifying which Public Key Signature Algorithm
+is supported by the MP JWT endpoint. This property can be be set to either `RS256` or `ES256`. Default value is `RS256`.
+
+`mp.jwt.verify.publickey.algorithm` will provide an additional hint how to read the Public Key (for example, EC Public Key in the PKCS#8 PEM format will only have a standard `-----BEGIN PUBLIC KEY-----` header and footer). It will also help avoiding the security risks associated with the token `alg`(algorithm) header dictating to the MP JWT endpoint which signature verification algorithm should be chosen.
```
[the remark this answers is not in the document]
rdebusscher: But this `alg` claim check is a breaking change, so ES256 support only in 2.0?

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -40,8 +40,10 @@ the MicroProfile services the token will be accessed with.
 The required minimum set of MP-JWT headers and claims is:
 
 typ:: This JOSE header parameter identifies the token as an RFC7519 and must be "JWT" https://tools.ietf.org/html/rfc7519#section-5.1[RFC7519, Section 5.1]
-alg:: This JOSE header parameter identifies the cryptographic algorithm used to secure the JWT. RSASSA-PKCS1-v1_5 SHA-256 algorithm is required for signing the claims and must be specified as "RS256", https://tools.ietf.org/html/rfc7515#section-4.1.1[RFC7515, Section 4.1.1]. RSAES using Optimal Asymmetric Encryption Padding and AES in Galois/Counter Mode (GCM) algorithms are required for
-encrypting the claims and must be specified as "RSA-OAEP", https://tools.ietf.org/html/rfc7518#section-4.3[RFC7518, Section 4.3] and "A256GCM", https://tools.ietf.org/html/rfc7518#section-5.3[RFC7518, Section 5.3]
+alg:: This JOSE header parameter identifies the cryptographic algorithm used to secure the JWT.
```
rdebusscher, line 43: Adding this is a breaking change, ES256 support only in 2.0?

api/src/main/java/org/eclipse/microprofile/jwt/config/Names.java
```diff
@@ -35,7 +35,12 @@
      * ClassLoader.getResource.  If the value is a URL it will be resolved using `new URL(“”).openStream()`
      */
     String VERIFIER_PUBLIC_KEY_LOCATION = "mp.jwt.verify.publickey.location";
-    
+
+    /**
+     * Public Key Signature Algorithm property which can be set to either 'RS256' or 'ES256'.
```
rdebusscher, line 40: Indicate the default value as RS256

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/SignatureAlgorithm.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ */
+package org.eclipse.microprofile.jwt.algorithm;
+
+/**
+ * JWT JSON Web Signature Algorithms which must be supported.
+ * 
+ * @see <a href="https://tools.ietf.org/html/rfc7518#section-3">https://tools.ietf.org/html/rfc7518#section-3</a>
+ */
+public enum SignatureAlgorithm {
```
rdebusscher, line 27: Only used in TCK, so should not be in API I guess

rdebusscher:
As the PR contains a breaking change, I would say we should forget about version 1.2 and put everything we have now in 2.0 as they would be released anyway around the same time (and only 2.0 will be included in MP 4.0)

spec/src/main/asciidoc/configuration.asciidoc
```diff
@@ -238,14 +251,23 @@ MicroProfile JWT implementations are required to throw a `DeploymentException` i
...
+
+MicroProfile JWT implementations are required to throw a `DeploymentException` if `mp.jwt.verify.publickey.algorithm` is not set to either `RS256` or `ES256`.
```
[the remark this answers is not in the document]
sberyozkin: @rdebusscher I thought your original concern was that MP JWT may start supporting every JWA algorithm. The implementations can easily support ES512 or RS512 etc if they need to with the implementation specific properties. IMHO the above restriction should be OK as `ES256` is the only other algorithm but `RS256` (and `HS256`) that I've seen being reported in the practical use.

Or do you prefer that we say `support for the other asymmetric signature algorithms is optional` ? (Given that there is a text there right now which prohibits `HS256`.
That can work too I guess

spec/src/main/asciidoc/configuration.asciidoc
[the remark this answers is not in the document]
sberyozkin: There is no any requirement here for the implementations to check `alg`manually - it is done by the underlying JOSE library. But what it says, now that 2 algorithms are supported, that using an `alg` check to determine which key to load is a security risk and the algorithm should be whitelisted before the verification starts - the point is the endpoint has a static upfront knowledge about the supported algorithm.
Also this property in itself does not change anything with respect to the way PEM keys have to be processed. Even now if someone puts an EC key into PEM then the implementations trying to get `RSAPublicKey` has to fail. Few things have to be tightened about the way the invalid keys are reported (recall we are having another discussion in relation to URL based keys, etc), but I'd rather prefer not to overload this PR. 
As such it is also a hint. If we have a PEM key with only `-----BEGIN PUBLIC KEY-----` then it can be either RSA or EC key. The implementations may want to use this property to optimize (ex, immediately read it as EC key as opposed to 1) read it as RSA key first and if it fails 2) as EC key). RSA PEM keys may have `RSA` qualifiers. So it is mainly a hint for reading the PEM keys. 

Yes, if we have a JWK set then this property provides for an additional `kty` check, apart from the matching `kid` requirement.

Let me try to simplify and rework the text a bit

spec/src/main/asciidoc/interoperability.asciidoc
sberyozkin: > Adding this is a breaking change, ES256 support only in 2.0?
It is not related to `ES256`. `alg` is already specified in MP JWT 1.1.1, I've just updated this text to clarify what this property contains when the token is encrypted.

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/SignatureAlgorithm.java
sberyozkin: @rdebusscher I'd prefer in our implementation to use a property shipped with MP JWT to initialize `mp.jwt.publickey.algorithm`. Other implementations may follow suit too for the same purpose or might want to use it as part of their implementations.

sberyozkin:
@rdebusscher 
> As the PR contains a breaking change
I've no idea what breaking change you refer to here.

> I would say we should forget about version 1.2 and put everything we have now in 2.0 as they would be released anyway around the same time (and only 2.0 will be included in MP 4.0)

No. MP JWT 1.2 will be in MP 4.0

sberyozkin:
@rdebusscher I've reviewed your comments where you suggested this PR can be `breaking`. I believe I've explained in my answers that no, it is not true. Let me summarize here though:
- `mp.jwt.publickey.algorithm` does not introduce any breaking changes for the users - it is a new property which is meant to help the implementations to deal with more than 1 supported algorithm.
- `alg` text in the `interoperability.adoc` - likewise, nothing breaking is introduced, the text has only been clarified what the `alg` will contain depending on how the token is secured. Nothing is breaking here.

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/SignatureAlgorithm.java
sberyozkin: @rdebusscher I can remove it altogether rather than move it to TCK. If you can not comfortable with it being shipped in the api artifact then I'll just drop it.

spec/src/main/asciidoc/configuration.asciidoc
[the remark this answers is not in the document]
rdebusscher: But now we have a conflicting text in the spec. In one place we say

>Other asymmetric signature algorithms are allowed, but should be considered vendor-specific.

And here we say

>are required to throw a `DeploymentException`

rdebusscher:
@sberyozkin 

About the `alg text in the interoperability.adoc`, you are correct. I have misread it before.

spec/src/main/asciidoc/interoperability.asciidoc
rdebusscher: I misread the diff, the addition doesn't seems to be a problem.

sberyozkin:
Hi @rdebusscher np, thanks for giving it a quick check, I'll also work on the text around the new property and try to avoid any ambiguities, etc

api/src/main/java/org/eclipse/microprofile/jwt/algorithm/SignatureAlgorithm.java
sberyozkin: @rdebusscher OK, probably simpler to drop it for now, I had it copied from the builder PR, but that PR is bigger and will not make it to 1.2. so I'll just drop this enum for now

spec/src/main/asciidoc/configuration.asciidoc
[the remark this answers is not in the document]
sberyozkin: @rdebusscher Yeah, I'd just drop the exception requirement

sberyozkin:
@rdebusscher I've tried to make it clearer how `mp.jwt.verify.publickey.algorithm` property should be used, moved `SignatureAlgorithm` to TCK as per you original proposal as it is used there and did few other minor updates. Have another check please. 
I don't expect that more than 1, max 2 days will be spent on implementing it. One would use `mp.jwt.verify.publickey.algorithm` to select between Java `RSA` and `EC` KeyFactory when loading a key, and set it as a whitelist property for Jose4J or other library and that is it really, plus few tests...

sberyozkin:
@rdebusscher Hi, are you happy now in general with the PR ? There will be time to tweak something related to this and indeed other 1.2 commits for sure before 1.2 is actually released.

sberyozkin:
@rdebusscher Hi Rudy, I've lifted the change request but I'll wait for another day or so in case you have something else in mind. As I said, we can tweak things easily as needed

sberyozkin:
@rdebusscher Going ahead with the merge - thanks for the review, please have a new issue opened if there will be some concerns in your team

PR #191, opened 2020-06-15
Simplify MP JWT header and claim requirements

### LINKED ISSUE #128: Need better clarify the Minimum MP-JWT Required Claims
MP-JWT defines a set of "Minimum MP-JWT Required Claims". But it does not explicitly specify what is the "required" means. Do we mean JWT must contain all claims? or MicroService must support this set of claims if they are included in JWT?  Maybe we should clarify which claims must be included, and which claim must be supported?
For example, if user does not belong to any group, do we want to accept the JWT? or if token replay is allowed, does JWT has to contain "jti"?

sberyozkin:
Fixes #128.

The goal of this PR is to improve/clarify the text around the required headers/claims and it will be followed up with a few more TCK tests once we agree on the final text.

Here is a summary of the problem. In the current text:
-  `alg` (and `enc` in case of the encrypted token), `kid ` and `typ` headers are required. 
- `iat`, `exp`, `iss`, one of `upn`/`preferred_username`/`sub`, `jti` claims are required in one place, `iss` - in the other place.

IMHO these are all very sound requirements but `required` must really be TCK enforced as it is important for some integrations that the spec requirements are enforced , but I suspect if we start enforcing it all  now we will break many user applications. So I've tried to find the middle ground. It is important that MP JWT has some position about some claims as opposed to making everything optional and accepting empty tokens. We can tighten things in MP JWT 2.0. So, in this PR:
 
1. Headers

- Required: `alg` and in case of the encrypted tokens also `enc` - this is essential and can not be optional
- Optional: `kid` - it is redundant unless JWK is used to verify the token. It really must be present  if JWK is used to verify but there is a border line case where only a single JWK is available without `kid` and can be used to verify the token without `kid` so I'd be a bit concerned with TCK enforcing that a `kid` header is present in such cases. Perhaps we can tighten it up for `MP JWT 2.0`
- Optional: `typ` it does not add any security information so it is a good practice to have it but I think it is fine to make it optional

2. Claims 
- `iat` and `exp` are still required - it simply does not make sense to accept the unlimited lifetime tokens (if no `exp` is available),  and also very risky to accept the tokens without `iat` (was it issued 5 years ago ? who knows...), every OIDC issuer will set it and it is straightforward to set for the self-issuing clients. Sometimes users don't want to set `exp` to make things easier to test, let then set a max long number, but at least with the `iat` requirement the implementations will have a chance to block the tokens which have been around forever... **There is no practical reason for a token not to have `iat` and/or `exp`**
- `iss` - this is required, without it the services simply have no idea who posted it, this is a fundamental JWT verification requirement. Again, every OIDC server will set it, and any self-issuing client can set it to something random.
- a principal must be available in the token (`upn`, or `preferred_username` or `sub`) - without it Java `Principal` `getName` will return nothing and the spec just can't allow it. Perhaps I should add a line suggesting the name can be optionally deduced from the other claims if none of these 3 claims is available.
- Optional: `sub` - (unless it is required as a name fallback), it is a number most of the time
- Opitonal: `jti` - this one should really be required but it can be enforced for `Mp JWT 2.0` with TCK, as I'm not 100% sure if all the OIDC servers set it (the self-issuing clients can set it without problems)

3. Missing Iss tests - I've removed them because I don't understand what they actually test, specifically the test where a token `iss` does not match a configured `issuer` seems just incorrect.

@rdebusscher, @starksm64, @radcortez @darranl, others, comments are welcome :-)


sberyozkin:
Or, another option, just leave it as is until `MP JWT 2.0` and TCK enforce the required claims then

sberyozkin:
I'm going to open for review since we need to agree on a decision quickly enough. As I said I will add a few more TCK updates once the text is agreed upon

sberyozkin:
@radcortez It is still not possible to request your review. Also CC @darranl, others

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -393,16 +404,19 @@ underlying implementation container.
...
+1. The JWT must have a JOSE "alg" header that indicates the token was signed using the RS256 or ES256 algorithm when the service endpoint expects signed tokens.
+2. The JWT must have the JOSE "alg" and "enc" headers that indicate that the token was encrypted using the RSA-OAEP and A256GCM algorithms when the service endpoint expects encrypted tokens.
+   If the encrypted content is a nested signed JWT token then it must meet the condition 1.
+3. The JWT must have an "iss" claim representing the token issuer that maps to an MP-JWT implementation container runtime configured value.
+   Any issuer other than those issuers that have been whitelisted by the container configuration must be rejected with an HTTP_UNAUTHENTICATED(401) error.
+4. The JWT must have an "iat" claim representing the token issuance time. It is impossible to calculate how long the token may have been used for without the "iss" claim so such tokens must rejected with an HTTP_UNAUTHENTICATED(401) error.
```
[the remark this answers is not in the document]
teddyjtorres: In addition, "must rejected" should read "must be rejected".

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -35,34 +35,45 @@ To meet those requirements, we introduce 2 new claims to the MP-JWT:
...
 jti:: Provides a unique identifier for the JWT. The identifier value MUST be assigned in a manner that ensures that
     there is a negligible probability that the same value will be accidentally assigned to a different data object;
     if the application uses multiple issuers, collisions MUST be prevented among values produced by different issuers
     as well.  The "jti" claim can be used to prevent the JWT from being replayed.  The "jti" value is a case-sensitive
     string. https://tools.ietf.org/html/rfc7519#section-4.1.7[RFC7519, Section 4.1.7]
-upn:: This MP-JWT custom claim is the user principal name in the java.security.Principal interface, and is the caller
-    principal name in javax.security.enterprise.identitystore.IdentityStore. If this claim is missing, fallback to
-    the http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims["preferred_username", OIDC Section 5.1] should be
-    attempted, and if that claim is missing, fallback to the "sub" claim should be used.
+aud:: Identifies the MP JWT endpoint(s) which can be accessed by JWT. https://tools.ietf.org/html/rfc7519#section-4.1.7[RFC7519, Section 4.1.3]
```
teddyjtorres, line 74: Fix link to https://tools.ietf.org/html/rfc7519#section-4.1.3

spec/src/main/asciidoc/interoperability.asciidoc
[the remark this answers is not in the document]
sberyozkin: @ayoho @teddyjtorres thanks for catching these typos, will be fixed shortly

sberyozkin:
Hi All, just to would to summarize again why some headers/claims are made optional with this PR:

`Headers`:

```
typ:: MP JWT only expects JWT. Given that JWT is basically just JSON where any top level element is a `claim`, having `typ` does not add to the security of the token - it is recommended to set this header to keep to keep indicating that well, it is not just random JSON but was created by the OIDC server or self-issuing client which did mean to issue JWT.
kid:: Mainly effective with the multi-key JWK set - and even in this case I'm not seeing the users requiring the use the certificate thumbprint `x5t` as an alternative 
```

`Claims`:
```
sub:: It is supposed to be a unique number so it is not something that would be useful to return to the human user as part of the UI interaction and it is really of interest to the OIDC server introspecting the token, not much can be done about it during the local verification. It is only required as a fallback when no `upn` or `preferred_username` is available 
jti:: hard to enforce it at the MP JWT spec level beyong just checking it exists - enforcing the replay constraints is better be left to the OIDC server or the custom local implementation. Perhaps we can do some general recommendations going forward.
aud:: I'd have no problems with enforcing it but setting it correctly is a bit harder for the self-issuing clients, and not all OIDC issued tokens may have them, so for now it is optional 
```

radcortez:
@sberyozkin I'm have permissions now, so I added myself as a reviewer :)

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -393,16 +404,19 @@ underlying implementation container.
...
+1. The JWT must have a JOSE "alg" header that indicates the token was signed using the RS256 or ES256 algorithm when the service endpoint expects signed tokens.
+2. The JWT must have the JOSE "alg" and "enc" headers that indicate that the token was encrypted using the RSA-OAEP and A256GCM algorithms when the service endpoint expects encrypted tokens.
+   If the encrypted content is a nested signed JWT token then it must meet the condition 1.
+3. The JWT must have an "iss" claim representing the token issuer that maps to an MP-JWT implementation container runtime configured value.
+   Any issuer other than those issuers that have been whitelisted by the container configuration must be rejected with an HTTP_UNAUTHENTICATED(401) error.
+4. The JWT must have an "iat" claim representing the token issuance time. It is impossible to calculate how long the token may have been used for without the "iat" claim so such tokens must be rejected with an HTTP_UNAUTHENTICATED(401) error.
```
sberyozkin, line 415: Fix the `4` duplication

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -393,16 +404,19 @@ underlying implementation container.
...
+4. The JWT must have an "exp" claim representing the token expiration time. Tokens without the "exp" claim have an unlimited lifetime and must be rejected with an HTTP_UNAUTHENTICATED(401) error.
+5. The token must have at least one of "upn" or "preferred_username" or "sub" claim for Java Principal to have a name. Tokens from which no Principal name can be deduced will be mapped to name-less Principals
```
radcortez, line 417: Maybe we can simplify?

> Tokens from which no Principal name can be deduced will be mapped to name-less Principals and therefore must be rejected with an HTTP_UNAUTHENTICATED(401) error.

To just:

> Tokens from which no Principal name can be deduced must be rejected with an HTTP_UNAUTHENTICATED(401) error.

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -35,34 +35,45 @@ To meet those requirements, we introduce 2 new claims to the MP-JWT:
...
     if the application uses multiple issuers, collisions MUST be prevented among values produced by different issuers
     as well.  The "jti" claim can be used to prevent the JWT from being replayed.  The "jti" value is a case-sensitive
     string. https://tools.ietf.org/html/rfc7519#section-4.1.7[RFC7519, Section 4.1.7]
-upn:: This MP-JWT custom claim is the user principal name in the java.security.Principal interface, and is the caller
-    principal name in javax.security.enterprise.identitystore.IdentityStore. If this claim is missing, fallback to
-    the http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims["preferred_username", OIDC Section 5.1] should be
-    attempted, and if that claim is missing, fallback to the "sub" claim should be used.
+aud:: Identifies the MP JWT endpoint(s) which can be accessed by JWT. https://tools.ietf.org/html/rfc7519#section-4.1.3[RFC7519, Section 4.1.3]
+
+[NOTE] MP JWT implementations may enforce that JWT tokens contain all the recommended headers and claims. The recommended headers and claims may become required in the future versions of the MP JWT specification.
```
radcortez, line 76: Does this mean that a MP JWT implementation may force for all recommended claims to be mandatory and another  to have them as optionals?

sberyozkin: @radcortez yes since `recommended` is `optional`.  The message here is that if a given implementation has an idea what to do about `jti` and the replay protection for example then it is encouraged to do it. Not sure, may be the language can become clearer :-)

radcortez: No, I think the text is clear, but may concern is that the JWT may not be portable across MP JWT impls. I'm thinking in scenarios where you have a JWT that is accepted on a service, propagates it to another service with a MP JWT implementation with stricter rules and then fail the JWT. I guess one way to fix it is to make the 2nd service happy, which should also work on the 1st.

On the other hand, since this is somehow open, you have have mix behaviour where some claims / headers are mandatory in an implementation and the other way around in another.

I'm not completely sure if we should leave this to the implementation. I see some potential issues here. What do you think?

spec/src/main/asciidoc/interoperability.asciidoc
sberyozkin: done

spec/src/main/asciidoc/interoperability.asciidoc
sberyozkin: This is like everything else where the spec lets the implementations do something optionally (example - some signature algorithms), if we leave the text out then it will be still the case with for example with the implementation X being relaxed about say the token age limit, the same token will be blocked by the downstream smallrye-jwt endpoint taking a harder stance :-).
The deployments involving potentially multiple vendor MP JWT impls would not be dynamic - the policy will have to be aligned. Which is what you suggested :-).

sberyozkin:
@radcortez @rdebusscher @teddyjtorres

I have added 3 TCK tests and this is probably enough for now.
- 401 is expected if no `exp` is available
- 401 is expected if `iat` is older than `exp` (Teddy, as per our discussion, before the age property is introduced)
- 401 is expected if the token has no name available

spec/src/main/asciidoc/interoperability.asciidoc
radcortez: If everyone else is ok with this, I'm also fine. Just wanted to raise the flag and clarify :)

sberyozkin:
Hey @radcortez, thanks, I'm going to verify TCK tests before merging.
@rdebusscher Rudy, any concerns - just ping us please :-).

sberyozkin:
The added TCK tests have been verified, merging shortly

PR #194, opened 2020-07-02
Add a TCK test for RSA keys of 1024 bit key length

### LINKED ISSUE #142: Update minimum key size to 2014
Section 9.2 of the specification states the required supported key sizes as: -

> Support for RSA Public Keys of 1024 or 2048 bits in length is required. Other key sizes are allowed,
> but should be considered vendor-specific.

Section 4.1 states "alg" must be "RS256": -

> This JOSE header parameter identifies the cryptographic algorithm used to secure the JWT. MPJWT requires the use of the RSASSA-PKCS1-v1_5 SHA-256 algorithm and must be specified as
> "RS256", RFC7515, Section 4.1.1

However RFC7518 "JSON Web Algorithms (JWA)" specifically mandates a minimum key size of 2048: -

> A key of size 2048 bits or larger MUST be used with these algorithms.

https://tools.ietf.org/html/rfc7518#section-3.3

I am raising this issue as I don't believe it would have been intentional to override the minimum supported key size.

### LINKED ISSUE #197: Add TCK test which confirms that JWTs signed with RSA keys of 1024 bit length are verified

sberyozkin:
Fixes #197

A simple test has been added to verify that a token signed with the 1024 bit length RSA key is accepted, it reuses the `RolesAllowed` endpoint

CC @darranl, @teddyjtorres

### DIFF
--- a/spec/src/main/asciidoc/configuration.asciidoc
+++ b/spec/src/main/asciidoc/configuration.asciidoc
@@ -47,9 +47,11 @@ HTTPS request.
 
 ### Supported Signature Algorithms
 
-Support for RSA RS256 and Elliptical Curve Digital Signature Algorithm (ECDSA) ES256 is required.
+Support for RSA RS256 and Elliptic Curve Digital Signature Algorithm (ECDSA) ES256 is required.
 RSA keys used for creating and verifying RS256 signatures must be of 1024 or 2048 bits in length.
 Other RSA key sizes are allowed, but should be considered vendor-specific.
+
+[NOTE] Support for RSA keys of 1024 bits in length is deprecated and will become optional in the next major version of the MP JWT specification.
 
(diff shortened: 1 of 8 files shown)


sberyozkin:
Hi All, I've updated this PR to fix the issue introduced with the migration to Jose4j - which is directly related to the RSA `1024` bit length. Jose4j enforces `2048+` length by default - which was not the case with `nimbus-jwt`. So TokenUtils.sign(...) now allows `1024` OOB.

sberyozkin:
TCK has been verified

sberyozkin:
@rdebusscher Hi Rudy, this is a simple PR so I'd like to go ahead asap to complete the 1.2 work, but I'll wait until tomorrow for you to provide the feedback. As always, please create a new issue if you'll get concerned with something about this PR in particular

PR #205, opened 2020-08-26
Remove 'build' from .gitignore

### LINKED ISSUE #192: Update gitignore 'build' exclusion
Hey @radcortez Are you working with Intellij by any chance ? `build` is excluded right now to eliminate some InteliJ build artifacts according to the comment, but it affects a package like `build`, I had to change from `build` to `builder` in the JWT generation PR.
Is it possible to make in more specific as opposed to just `build` ? Not urgent or critical...

sberyozkin:
Fixes #192

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@
 target/
 tck/bin/
 .project
-build/
 .classpath
 .factorypath
 test-output


rdebusscher:
Any reason why the contents of build directories needs to be in the git repo?

sberyozkin:
@rdebusscher Hi, see #192

It is not a big deal that for example a future builder API would have `builder` package as opposed to `build` (the latter is more compact) but keeping `build` ignored just in case is not worth on its own, neither Eclipse nor IntelliJ nor Maven apparently have this folder generated. 
I've no strong opinion but I'd like to see someone actually saying, wait. my IDE generates `build`.
It was added here:
https://github.com/eclipse/microprofile-jwt-auth/commit/792c58104ec2fa51948e5621177b0426d0ddbba4#diff-a084b794bc0759e7a6b77810e01874f2

Let me ask @starksm64 Hi Scott, was there any particular reason `build` was added in `.gitignore` ?

My best bet is that it is the Ant legacy and `build` is a copy and paste side-effect, but I won't rush

sberyozkin:
For the record, smallrye-jwt JWT builder API which the build JWT PR was based around has `build` but we haven't seen anyone reporting any issues around it in the Quarkus/smallrye-jwt land for the last 6+ months.

rdebusscher:
Then there was maybe a typo in the .gitignore initially when ant build directory was meant to exclude (should be `/build` instead of `build/`)

radcortez (approved):

Emily-Jiang (approved):

Issue #182, opened 2020-04-13
Drop @LoginConfig


rmannibucau:
Follow up issue of https://github.com/eclipse/microprofile-jwt-auth/issues/70 (hoping the scope of this one is clear and narrowed :))

This annotation has a lot of pitfalls:

1. it is not defined in a way which enables it to be portable,
2. it requires to hardcode the security config in the application (same as @*Definition are never used anywhere else than in demo in JakartaEE land, we should learn from this error),
3. it is not needed to make it work (only case it can be relevant is for a container providing all spec, even the ones the user don't need/want - container philosophy - and in this case the vendor can trivially provide a way to disable it if needed so the toggle is not in jwt-auth spec level).

So overall, there is not a single case which justify this annotation and it does not bring any configuration the implementation would require so there is no blocker to drop it (note that dropping an annotation keeps the application functional without recompilation in terms of API).

Side note: adding a mp-config key to enable/disable this spec can still be done as a consequence of the removal of this annotation IMHO.

sberyozkin:
@rmannibucau Sure, this is more concrete, no doubt about it :-)

sberyozkin:
Hey @rmannibucau I've added a 2.0 milestone, though I'll resist as much as possible :-) because we have the users (and likely not only us) using it, unfortunately with more than MP-JWT, so there has to be a strong reason to make them update their applications. Few quick questions:

> it is not defined in a way which enables it to be portable,

What is not portable about sticking it to the JAX-RS `Application` ?

> it requires to hardcode the security config in the application 

Do you mean its optional `realm` property ?

> it is not needed to make it work

MP-JWT ? I certainly know one project where `authMethod` is checked.

I wonder if a starting point can be be to drop the references to all but `MP-JWT` method, and make `MP-JWT` a default value. IMHO it can be more realistic than just dropping `LoginConfig`

Cheers

rmannibucau:
> there has to be a strong reason to make them update their applications

It is an annotation so if not in the classpath it is ignored, no update to do :).

> What is not portable about sticking it to the JAX-RS Application ?

It is not defined yet ;). It also has the issue to conflict with other EE specs (servlet and security out of my mind) which have their own security mecanism/annotations. Who wins? Who wraps who?

> Do you mean its optional realm property ?

[...]

PR #206, opened 2020-08-26
LoginConfig: remove refs to all but the MP-JWT method and make MP-JWT the default one

sberyozkin:
Related #182 

This PR is the first step in the long (possibly never ending) process of dropping `LoginConfig`:
- removes the references to all but `MP-JWT` methods - `MP-JWT` spec should really be only about JWT; note dropping these references from the doc only serves this purpose, i.e, stop sending the message that `LoginConfig` can be an annotation for all sort of auth mechanisms
- `MP-JWT` is a default method now - it does not break the backward compatibility since until now setting the auth method is required, I suspect in the vast majority of cases where this annotation is still used it is `@LoginConfig(authMethod="MP-JWT")` so at least the users will now be able to do just `@LoginConfig`...
- updated the baseline version in this package to `1.1` - I'm scared of this update the most :-), @Emily-Jiang, please check if it is correct

I really don't think we can do more about `@LoginConfig` in the short/medium term
Hi Romain, @rmannibucau, not sure why but could not add you as a reviewer, I'm doing this PR given your last comment at #182 
Also see @teddyjtorres

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/auth/LoginConfig.java
+++ b/api/src/main/java/org/eclipse/microprofile/auth/LoginConfig.java
@@ -27,8 +27,10 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * A security annotation describing the authentication method, and the associated realm name that
+ * A security annotation describing the authentication method and the associated realm name that
  * should be used for this application.
+ *
+ * Note: this annotation may be removed in the future versions of the MP JWT specification.
  *
  */
 @Inherited
@@ -37,27 +39,19 @@ import java.lang.annotation.Target;
 @Retention(RetentionPolicy.RUNTIME)
 public @interface LoginConfig {
     /**
-     * Soften the requirements here:
-     *
-     * The authMethod is used to configure the authentication mechanism for the
+     * The 'authMethod' is used to configure the "MP-JWT" authentication mechanism for the
      * JAX-RS application. As a prerequisite to gaining access to any web resources
      * which are protected by an authorization constraint, a user must have
-     * authenticated using the configured mechanism. Supported values include
-     * "BASIC", "DIGEST", "FORM", "CLIENT-CERT", "MP-JWT", or a vendor-specific
-     * authentication scheme.
-     *
-     * Note the the MP-JWT TCK currently only validates that a deployment with
-     * MP-JWT authentication follows the specification, but in the future,
-     * we MAY look to combine the use of MP-JWT tokens with other authentication
-     * mechanisms.
+     * authenticated using the configured mechanism.
+     * Support for the other authentication mechanisms is optional.
      *
      * @return the configured auth-method
      */
     String authMethod();
 
     /**
-     * The realm name element specifies the realm name to
-     use in HTTP Basic authorization.
+     * The realm name
+     *
      * @return the realm name which may be empty
      */
     String realmName() default "";


rmannibucau:
@sberyozkin think this change does not need to go with a 1.1 change but ok for me as a move in the right direction.

Side question related to this issue plannification: when is mp-jwt-auth 2 planned? Cause makes long enough without any change to drop the annotation and do a 2.0 now (most MP spec broke often so guess jwt auth can too now ;)).

sberyozkin:
Hi @rmannibucau for MP JWT it would be `1.2` which will go into MP 4.0, the process is slow right now so i can only guess MP JWT 2.0 with the breaking changes if any would be in 2021. MP JWT 1.2 has a good number of various non-breaking updates and I could not see/find the scope for going for more and doing a proper 2.0.

rmannibucau:
@sberyozkin ok, just thought that cause of all the pain other specs did with unexpected breaking changes this one could go but also agree it is just a dead annotation so while doc is clear it is already a very good move. Thanks Sergey.

rmannibucau (approved):

Emily-Jiang:
@sberyozkin don't need to be scared with the version changes. The bnd plugin will work it out for you. When you do `mvn verify` and it builds cleanly, you are good. Otherwise, the plugin will tell you what to do. Your changes look good to me.

Emily-Jiang (approved):

api/src/main/java/org/eclipse/microprofile/auth/LoginConfig.java
```diff
@@ -37,27 +37,19 @@
 @Retention(RetentionPolicy.RUNTIME)
 public @interface LoginConfig {
     /**
-     * Soften the requirements here:
-     *
      * The authMethod is used to configure the authentication mechanism for the
      * JAX-RS application. As a prerequisite to gaining access to any web resources
      * which are protected by an authorization constraint, a user must have
-     * authenticated using the configured mechanism. Supported values include
-     * "BASIC", "DIGEST", "FORM", "CLIENT-CERT", "MP-JWT", or a vendor-specific
```
Emily-Jiang, line 46: Why did you delete these?

sberyozkin: @Emily-Jiang The idea is stop start preparing for the eventual removal of this annotation; there was a long discussion in of the old issues with many people preferring it; but it can also be a very sensitve thing to do in some cases, so what we agreed with Romain is to start with dropping any references to all but `MP-JWT` authentication mechanism for the users not to keep thinking that the MP JWT API introduces the annotation which can be used even for non-MP-JWT flows (which was one of the problems discussed in the earlier discussion), the scope of the MP JWT should be all about MP JWT.

sberyozkin:
Hi @Emily-Jiang thanks, yes, it failed during the local build and suggested `1.0.1` but I thought aligning it with `1.1` would be better.
Having said it, I've been thinking about the whole default `authMethod` enhancement and thought I made a pointless enhancement really as `LoginConfig` on its own does not give any indication at all of what this `LoginConfig` is about and in a way it contradicts with our plan to deprecate/remove this annotation as it makes it easier to use it for the `MP-JWT` auth mechanism.
So I'm going to revert the code/package changes and only trim `LoginConfig` docs for `MP-JWT` only be referenced

sberyozkin:
Hi All, so now it is only the docs update keeping a reference to only `MP-JWT` and sending a message with `Support for the other authentication mechanisms is optional.` that it is really not the way to go with using this `MP-JWT` annotation for all the auth methods out there. The `optional` bit is not breaking I think as TCK does not enforce this annotation is effective anyway. If someone is concerned about `optional` then I will change
`Support for the other authentication mechanisms is optional.` to for example `Other authentication mechanisms may be supported` which is the same really.
But in any case removing the refs to `FORM`/etc cleans it up and will make it easier for us to drop this annotation later on...

rdebusscher:
Should we not mention already that it is the idea to remove the annotation all together later?  It will trigger maybe people to give some scenarios which we are not thinking about today but are important.

api/src/main/java/org/eclipse/microprofile/auth/LoginConfig.java
Emily-Jiang: ok, thanks

radcortez (approved):

sberyozkin:
@rdebusscher Hi, yes, makes sense, will update with the Java docs with something like "Note: this annotation may be removed in the future versions of the MP JWT specification". I suppose adding `@Deprecated` would also be the right thing to do, but with `@Deprecated` it would be good to have an alternative link so I'll probably avoid it.
By the way, is there some similar annotation in Jakarta Security ? I've looked around but could not see anything...Also CC @teddyjtorres

sberyozkin:
@radcortez thanks, I'll do another small update to note the annotation may be removed, and then it should be ready for a merge

sberyozkin:
@rdebusscher So I've added a note that this annotation can be removed which will let some users ask the questions if they need to before it is probably gone in 2.0/etc.

sberyozkin:
Going with merge now, any concerns - please open a new issue

PR #210, opened 2020-09-23
Issue 209: Add missing verifyTokenWithoutName endpoint

### LINKED ISSUE #209: 404 reported in verifyTokenWithoutName because test endpoint does not exist
Running the TCK against my product shows the following failure in the [org.eclipse.microprofile.jwt.tck.container.jaxrs.RequiredClaimsTest.verifyTokenWithoutName](https://github.com/eclipse/microprofile-jwt-auth/blob/f2a0557d1ec0544b9c404216ff847dc401814fb4/tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/RequiredClaimsTest.java#L283) test:

```
expected [401] but found [404]

java.lang.AssertionError: expected [401] but found [404]
at org.testng.Assert.fail(Assert.java:94)
at org.testng.Assert.failNotEquals(Assert.java:496)
at org.testng.Assert.assertEquals(Assert.java:125)
at org.testng.Assert.assertEquals(Assert.java:372)
at org.testng.Assert.assertEquals(Assert.java:382)
at org.eclipse.microprofile.jwt.tck.container.jaxrs.RequiredClaimsTest.verifyTokenWithoutName(RequiredClaimsTest.java:291)
```

I believe this is a bug in the test application used by the TCK. The problem is that the endpoint class (https://github.com/eclipse/microprofile-jwt-auth/blob/f2a0557d1ec0544b9c404216ff847dc401814fb4/tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/RequiredClaimsEndpoint.java) that defines the endpoints invoked by the tests doesn’t have a “/verifyTokenWithoutName” endpoint defined, while all of the other tests do.

ayoho:
Resolves #209

Adds the `/verifyTokenWithoutName` endpoint to the `RequiredClaimsEndpoint` class to avoid encountering a 404 by the test.

Signed-off-by: Adam Yoho <ayoho@us.ibm.com>

### DIFF
--- a/tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/RequiredClaimsEndpoint.java
+++ b/tck/src/test/java/org/eclipse/microprofile/jwt/tck/container/jaxrs/RequiredClaimsEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2017 Contributors to the Eclipse Foundation
+ * Copyright (c) 2016-2020 Contributors to the Eclipse Foundation
  *
  *  See the NOTICE file(s) distributed with this work for additional
  *  information regarding copyright ownership.
@@ -271,6 +271,12 @@ public class RequiredClaimsEndpoint {
     }
 
     @GET
+    @Path("/verifyTokenWithoutName")
+    public Response verifyTokenWithoutName() {
+        throw new RuntimeException("This method must not be invoked");
+    }
+
+    @GET
     @Path("/verifyTokenWithoutClaim")
     public Response verifyTokenWithoutClaim() {
         throw new RuntimeException("This method must not be invoked");
@@ -282,4 +288,3 @@ public class RequiredClaimsEndpoint {
         throw new RuntimeException("This method must not be invoked");
     }
 }
-


sberyozkin:
@ayoho Thanks for this PR, can you please sign `eclipsefdn/eca ` ?

@radcortez Hi Roberto, looks like the latest snapshots are not picked up by our git hub actions in smallrye ? We have all the tests passing but we should have one failing before Adam's PR gets applied :-)

radcortez:
Ah, yes. We are pointing to the RC release: https://github.com/smallrye/smallrye-jwt/blob/mpjwt12/pom.xml#L38. I'll update to point to the SNAPSHOT.

sberyozkin:
Hi @ayoho we've discussed it with Roberto, I think the reason we don't see the test failing is that we have a verification done before the JAX-RS method resolution is done, i.e, it is JAX-RS Pre-Match. But I think having this guard would be fine as well to support different implementations

sberyozkin:
@ayoho Hi Adam, if you'll have problems signing it in then we can merge it later, but indeed if you could get the commits coming then it would be great :-)

ayoho:
@sberyozkin Hi Sergey, sorry about the delay - I had signed the agreement but hadn't done the re-validation of the issue yet. Should be good now 👍

sberyozkin (approved):
@ayoho np, sorry for my delay as I had to temp sign off, thanks for the PR

PR #216, opened 2020-10-22
Use signed+encrypted tokens for PrivateKeyAsX tests

ayoho:
Signed-off-by: Adam Yoho <ayoho@us.ibm.com>


sberyozkin:
Thanks @ayoho I'm going to check it and will merge asap

sberyozkin:
Some context for everyone: these tests have been using the encrypted-only tokens which contradicts to the fact that the encryption-only is optional

tck/src/test/java/org/eclipse/microprofile/jwt/tck/config/jwe/PrivateKeyAsJWKClasspathTest.java
```diff
@@ -103,9 +107,10 @@ public static WebArchive createLocationDeployment() throws IOException {
     public void testKeyAsLocation() throws Exception {
         Reporter.log("testKeyAsLocation, expect HTTP_OK");
 
+        PrivateKey signingKey = TokenUtils.readPrivateKey("/privateKey4k.pem");
         PublicKey publicKey = TokenUtils.readJwkPublicKey("/encryptorPublicKey.jwk");
         String kid = "mp-jwt";
-        String token = TokenUtils.encryptClaims(publicKey, kid, "/Token1.json");
+        String token = TokenUtils.signEncryptClaims(signingKey, null, publicKey, kid, "/Token1.json", false);
```
sberyozkin, line 113: @ayoho the last parameter has to be `true` - otherwise the test must fail because with the inner-signed tokens being encrypted, the JWE `cty` header must be set to `JWT` - this is clarified in the spec I believe and the existing sign/encrypt TCK tests enforce it, see also
https://github.com/eclipse/microprofile-jwt-auth/blob/master/tck/src/main/java/org/eclipse/microprofile/jwt/tck/util/TokenUtils.java#L452
thanks

ayoho: You're right, I had missed that part of the spec. I've pushed a new commit that changes these values to `true`.

sberyozkin: @ayoho thanks, it looks good. Can you please squash your 2 commits ?

sberyozkin:
@ayoho Not sure why the eclipsefdn/eca has failed now, all was good for your previous PRs since you signed it. I can merge, but please investigate

sberyozkin:
@ayoho Can you please remove [this test entry](https://github.com/eclipse/microprofile-jwt-auth/blob/master/tck/src/test/resources/suites/tck-base-suite.xml#L80) plus the supporting test/endpoint code as well since it also contradicts to the fact the encryption-only is optional ? I should've updated the spec to require supporting the encryption-only for the cookie cases (signing part only adds an overhead there), but it is too late now, can be reviewed for 2.0 though.
Thanks

ayoho:
@sberyozkin Done - I've removed the test class, the `microprofile-config-decrypt.properties` file that was used only by that class, and removed the class entry from the TCK suite. Looks like the eclipsefdn/eca check is also better now.

sberyozkin (approved):

PR #217, opened 2020-11-01
Update TCK to depend on MP Config 1.4

sberyozkin:
Hi Roberto, this is a minor update, given that you spotted 1.3 was too old; other versions can be considered but for now this is a safe update :-)

### DIFF
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -29,6 +29,7 @@
     <description>Eclipse MicroProfile JWT Feature - TCK</description>
 
     <properties>
+        <version.microprofile.config>1.4</version.microprofile.config>
         <version.shrinkwrap.resolvers>2.2.6</version.shrinkwrap.resolvers>
         <version.testng>6.10</version.testng>
         <version.jose4j>0.7.0</version.jose4j>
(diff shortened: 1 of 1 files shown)


tck/pom.xml
```diff
@@ -29,6 +29,7 @@
     <description>Eclipse MicroProfile JWT Feature - TCK</description>
 
     <properties>
+        <version.microprofile.config>1.4</version.microprofile.config>
```
Emily-Jiang, line 32: I think you will need to depend on Config 2.0 since JWT 1.2 and Config 2.0 are under MP 4.0.

radcortez: Yes, this was just until we get the proper MP Config 2.0 release.

radcortez (approved):

tck/pom.xml
sberyozkin: @Emily-Jiang Yeah, we are waiting for MP Config 2.0, but this is just a minor update for us to check 1.4 (as it should've been set earlier as spotted bu Roberto) is fine as expected, thanks

PR #221, opened 2020-11-17
Add the build-helper-maven-plugin to attach spec artifacts

starksm64:

Addresses the issue I mentioned at https://groups.google.com/u/1/g/microprofile/c/T_XcmTbQp6Q where the
staged org.eclipse.microprofile.jwt:microprofile-jwt-auth-spec artifact is missing the html/pdf spec contents.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
       - uses: actions/checkout@v2
         name: checkout
 
-      - uses: actions/setup-java@v1.3.0
+      - uses: actions/setup-java@v1.4.3
         name: set up jdk ${{matrix.java}}
         with:
           java-version: ${{matrix.java}}
--- a/spec/pom.xml
+++ b/spec/pom.xml
@@ -35,6 +35,8 @@
         <license>Apache License v 2.0</license>
         <maven.build.timestamp.format>MMMM dd, yyyy</maven.build.timestamp.format>
         <revisiondate>${maven.build.timestamp}</revisiondate>
+        <config.spec.pdf>${project.build.directory}/generated-docs/${project.build.finalName}.pdf</config.spec.pdf>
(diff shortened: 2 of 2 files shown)


sberyozkin:
Hi @starksm64 thanks, LGTM, do we have to test it with `1.2-RC3` ?

sberyozkin:
Hmm, the builds are failing, @radcortez hey Roberto, looks like the github action issue ?

starksm64:
I updated the setup-java action to 1.4.3 to address the build failure. Let's see if that fixes it.

starksm64:
@sberyozkin I was going to do a 1.2.2-RC1 release to continue testing the promotion CI job I'm testing and that will test this. I did not want to affect the main 1.2 releases for this testing.

sberyozkin:
Forgot to approve via the Review Action, so for the record: Approved

sberyozkin:
Hi Scott @starksm64, I guess if the next maintenance release, after `1.2` would be for example `1.3` then it is fine, for `MP 4.1`, if it happens; if MP-JWT maintenance version will have to stay in the `1.2.x` line, then again, if it is `1.2.1` then all is good; but what if we end up going into `1.2.2` ? Not 100% sure what would happen if for ex `1.2.2-RC1` were already available ? (we can start in that case from `1.2.2-RC2` etc as needed which is probably fine too)

I'd have no problems with `1.2-RC3`, `1.2-RC4` :-) but if you feel it is better not to touch the main releases for your testing then it is fine
thanks

radcortez:
> I updated the setup-java action to 1.4.3 to address the build failure. Let's see if that fixes it.

Yes, GH removed an API used by version 1.3, so it is failing now. To be on the safe side we could just point to `actions/setup-java@v1`. It will update minor and fix versions of the plugin automatically

PR #222, opened 2020-11-17
Correct the config.spec.* variables to not include the version.

starksm64:
Running an install did not work. This fixes the problem with the config.spec.pdf/config.spec.html settings.

### DIFF
--- a/spec/pom.xml
+++ b/spec/pom.xml
@@ -35,8 +35,8 @@
         <license>Apache License v 2.0</license>
         <maven.build.timestamp.format>MMMM dd, yyyy</maven.build.timestamp.format>
         <revisiondate>${maven.build.timestamp}</revisiondate>
-        <config.spec.pdf>${project.build.directory}/generated-docs/${project.build.finalName}.pdf</config.spec.pdf>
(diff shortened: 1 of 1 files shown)


sberyozkin (approved):

spec/pom.xml
```diff
@@ -35,8 +35,8 @@
         <license>Apache License v 2.0</license>
         <maven.build.timestamp.format>MMMM dd, yyyy</maven.build.timestamp.format>
         <revisiondate>${maven.build.timestamp}</revisiondate>
-        <config.spec.pdf>${project.build.directory}/generated-docs/${project.build.finalName}.pdf</config.spec.pdf>
-        <config.spec.html>${project.build.directory}/generated-docs/${project.build.finalName}.html</config.spec.html>
+        <config.spec.pdf>${project.build.directory}/generated-docs/${project.artifactId}.pdf</config.spec.pdf>
```
Emily-Jiang, line 38: @starksm64 during the MP meeting, we decided to include the version so that when people download the doc, they know which version of the spec they are reading. I think you need to revert the change.

PR #230, opened 2020-12-16
Fixed typo using an indefinite article for vocals

commodis:


commodis:
I am waiting for eclipse account creation

rdebusscher:
But a URL is valid, https://whatis.techtarget.com/feature/Which-is-correct-a-URL-or-an-URL

commodis:
I guess so, thus closing. Thanks for the review

PR #236, opened 2021-01-06
Minor JavaDoc and specification fixes

### LINKED ISSUE #234: Spec Bug under "Additional Claims"
In the "Additional Claims" paragraph (p17 of the PDF) the specification contains a bug because it mentions a `JsonWebToken getOtherClaim(String)` method that neither exists in the code extracts above it nor in the latest JavaDoc.

### LINKED ISSUE #235: Incomplete JavaDoc in JsonWebToken
The JavaDoc of `JsonWebToken` is incomplete at least in the `getSub()` method:
"This is the token issuing IDP subject, not the"
 is an incomplete sentence, there is nothing after "not the" explaining what it isn't.

sberyozkin:
Fixes #234.
Fixes #235.

@keilw has reported these 2 issues and this PR fixes them. I don't think it requires a `1.2.1` maintenance release, these are the obvious and old typo/text bugs, no-one has noticed it during the last few years, so I'd not make a big deal out of it - it will be in the next platform release anyway. That said, I don't mind much doing a `1.2.1` tag.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
+++ b/api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
@@ -70,8 +70,7 @@ public interface JsonWebToken extends Principal {
 
     /**
      * The sub(Subject) claim identifies the principal that is the subject of
-     * the JWT. This is the token issuing
-     * IDP subject, not the
+     * the JWT. This is the token issuing IDP subject.
      *
      * @return the sub claim.
      */
--- a/spec/src/main/asciidoc/interoperability.asciidoc
+++ b/spec/src/main/asciidoc/interoperability.asciidoc
@@ -146,8 +146,7 @@ public interface JsonWebToken extends Principal {
 
     /**
      * The sub(Subject) claim identifies the principal that is the subject of
-     * the JWT. This is the token issuing
-     * IDP subject, not the
+     * the JWT. This is the token issuing IDP subject.
      *
      * @return the sub claim.
      */
@@ -247,7 +246,7 @@ public interface JsonWebToken extends Principal {
 
 ### Additional Claims
 The JWT can contain any number of other custom and standard claims, and these are
-made available from the JsonWebToken getOtherClaim(String) method. An example
+made available from the JsonWebToken getClaim(String) method. An example
 MP-JWT that contains additional "auth_time", "preferred_username", "acr",
 "nbf", "aud" and "roles" claims is:
 


radcortez (approved):

sberyozkin:
Some Maven dependency pull issue, might take a few runs to pass

keilw:
I will take care of the much wider-spread Jakarta EE / Java EE inconsistencies and outdated stuff. 
While the spec update is clearer, one thing it does not fully explain, what if you enter a bogus claim name like "foo" do you get a null or "empty" `Optional` then or an exception?

sberyozkin:
@keilw 

> I will take care of the much wider-spread Jakarta EE / Java EE inconsistencies and outdated stuff.

Sure, sounds good - note I'd appreciate that whatever we do at this stage at the spec level remains a minor editorial update; we'd need to coordinate with the rest of MP specs in how we approach JakartaEE level updates for 2.0.

> one thing it does not fully explain, what if you enter a bogus claim name like "foo" do you get a null or "empty" Optional then or an exception?

It is not a bogus claim - this is just a custom claim name which may or may nor exist. See JavaDocs, it  [will be null](https://github.com/eclipse/microprofile-jwt-auth/blob/master/api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java#L146).

sberyozkin:
@radcortez @keilw @ederks85 @Emily-Jiang Hey, I'll start a discussion next week about 1.2.1

keilw:
@sberyozkin You mean "foo" would be a custom claim, well according to JWT anything would be possible ;-)

As at least the Annotation spec: https://jakarta.ee/specifications/annotations/1.3/annotations-spec-1.3.html in 1.x never seems to have been completely transferred over properly until 2.x that is the only place where referring to at least the JSR documents is the only way, but it still should be called Jakarta EE there because it is.

sberyozkin:
@keilw 

> You mean "foo" would be a custom claim, well according to JWT anything would be possible ;-)

Is it a question or assertion ? Yes, a token may have not only standard but custom claims, these claims can be simple or complex, and can have whatever names the users have decided to name them with

> As at least the Annotation spec: https://jakarta.ee/specifications/annotations/1.3/annotations-spec-1.3.html in 1.x never seems to have been completely transferred over properly until 2.x that is the only place where referring to at least the JSR documents is the only way, but it still should be called Jakarta EE there because it is.

I don't it is relevant to this PR

thanks

PR #238, opened 2021-02-19
Minor specification fixes

jumapico:
Fix package and import of LoginConfig


sberyozkin (approved):
@jumapico thanks for your PR. Please note that `LoginConfig` may get removed in `2.0`, depending on the resolution of the issue requesting its deletion.

PR #247, opened 2021-07-14
Add Tests Ran in Base Suite to Full Suite

Pandrex247:
It appears the full-suite xml has fallen behind the base-suite and runs substantially fewer tests.

### DIFF
--- a/tck/src/test/resources/suites/tck-full-suite.xml
+++ b/tck/src/test/resources/suites/tck-full-suite.xml
@@ -28,6 +28,7 @@
                 <include name="cdi" description="Base CDI injection of ClaimValues"/>
                 <include name="cdi-json" description="CDI injection of JSON-P values"/>
                 <include name="cdi-provider" description="CDI injection of javax.inject.Provider values"/>
+                <include name="config" description="Validate configuration using MP-config"/>
             </define>
             <define name="excludes">
                 <include name="utils-extra" description="Additional utility tests" />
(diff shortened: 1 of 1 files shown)


sberyozkin:
@Pandrex247 Thanks for this PR. I think the reason this resource has not been looked at in detail yet is because there is no guideline as it what does it mean to pass a base TCK suite vs the full one.
By the way is it possible to somehow include the base TCK tests by extending the base resource ?

sberyozkin:
I've assigned it to an `MP JWT 3.0` milestone - since the guideline for 2.0 in only include the TCK test or doc fixes if any, in addition to the Jakarta namespace update.
The fix is safe but we'd also need to update the spec with the full vs base TCK comparison hence I think it is not an `MP JWT 2.0` material

Pandrex247:
The definition of base vs. full is described here:
https://github.com/eclipse/microprofile-jwt-auth/tree/master/tck#optional-java-ee-container-integration-tests

> In addition to the categories of tests defined above, there are optional tests that illustrate the expected container API integration for Java EE container runtimes beyond JAX-RS. These tests are marked with the TestNG groups "ejb-optional", "jacc-optional", "servlet-optional" and "ee-security-optional". An MP-JWT implementation is not required to pass these tests in order to be considered a valid implementation. The TestNG testsuite for these categories + the base categories of tests can be found in the tck directory src/test/resources/suites, file tck-full-suite.xml.

So by that description the "full" suite should run the base and extra optional tests.

I'm not sure ottomh about extending XML configs, you can chain them though. So base would just contain base and "full" would just contain the extra optional tests, and you'd pass TestNG both files when running it.

sberyozkin:
Hi @Pandrex247 Sure, what I've meant is that at the MP level we do not have a differentiation between a basic and a full compliance.

sberyozkin:
@Pandrex247 Your PR should be good to go to Mp JWT 2.0 - the full suite file is indeed broken right now - at least should have it in order, thanks

sberyozkin (approved):

PR #249, opened 2021-07-28
Bump MicroProfile Parent version to 2.1

sberyozkin:


sberyozkin:
The build will fail, but I'll continue with renaming next

radcortez (approved):

Issue #252, opened 2021-08-26
Release this spec to be part of 5.0


Emily-Jiang:
- [x] Create release note 2.0 for the spec
- [x] Perform a release 2.0-RC1 off the branch jakarta91-branch to be included in MP 5.0

Emily-Jiang:
@sberyozkin can you resolve this issue asap? Most of the specs have created release candidates. Please let me know if you need any help with this.

sberyozkin:
@Emily-Jiang, can you clarify please, 
Do we really need the release note for `2.0` while we are releasing `2.0-RC1` ?
And what do you mean by `Perform a release 2.0-RC1 off the branch jakarta91-branch to be included in MP 5.0` - MP 5.0 will include the final `2.0` release ?
Also #257 has to be fixed for 2.0 - where do we record it, in this task ?

Emily-Jiang:
> @Emily-Jiang, can you clarify please,
> Do we really need the release note for `2.0` while we are releasing `2.0-RC1` ?
Yes, you need a release note in the spec. The release note is for releasing 2.0. `2.0-RC1` is the release candidate for 2.0. You should mention the 2.0 release. See an [example](https://download.eclipse.org/microprofile/staging/microprofile-config-3.0-RC4/microprofile-config-spec-3.0-RC4.html#release_notes_30) .
> And what do you mean by `Perform a release 2.0-RC1 off the branch jakarta91-branch to be included in MP 5.0` - MP 5.0 will include the final `2.0` release ?
> Also #257 has to be fixed for 2.0 - where do we record it, in this task ?
No need to mention #257. There is no user impact what so ever. Besides, it is just a check sytle, part of the consuming parent pom formatter. If you want, you can associate with the issue of consuming parent pom.xml. I am okay with the new style. If you want to fix it back, just put in the 2.0 milestone.

sberyozkin:
@Emily-Jiang I think #257 will have to be fixed for 2.0 - it is just too messy to have a build producing a new format every time and it looks not good IMHO. I'll talk to Roberto about it. But for RC1 it is not a problem.

sberyozkin:
@Emily-Jiang The staging repo is here, https://oss.sonatype.org/content/repositories/orgeclipsemicroprofile-1507, will promote shortly

sberyozkin:
[...]

PR #255, opened 2021-09-01
Jakarta9 source updates

sberyozkin:


sberyozkin:
@Emily-Jiang @radcortez never mind, I haven't rebased the branch

sberyozkin:
@Emily-Jiang @radcortez Something has changed in the `2.1` parent so now `// @formatter off` in `api/src/main/java/org/eclipse/microprofile/jwt/Claims.java` is ignored and it is misformatted. Do you know what it can be ?

Emily-Jiang:
[Here](https://github.com/eclipse/microprofile-parent/commits/master) are the changes in microprofile-parent

sberyozkin:
Hi @Emily-Jiang - if you have some time, can you please checkout this PR's branch - and independently verify `Claims.java` is affected ?
Please also note I had to revert some TCK source file changes:
https://github.com/sberyozkin/microprofile-jwt-auth/commit/7eec086a1d84889dfc2a4adab3052d48ad214eb5
(EJB, crypto, servlet classes and possibly a few others have no `jakarta.*` equivalents) - not sure if it is parent problem or something else
I think the package.info update (see the last commit) is fine, but check please

thanks

sberyozkin:
Note in https://github.com/sberyozkin/microprofile-jwt-auth/commit/7eec086a1d84889dfc2a4adab3052d48ad214eb5 the package import order is changed again after building the TCK, so the actual fixes `from jakarta.* back to javax.*` are not very easy to see

spec/src/main/asciidoc/interoperability.asciidoc
```diff
@@ -676,36 +676,36 @@ In terms of mapping between the MP-JWT claims and role names used in @RolesAllow
 ### Recommendations for Optional Container Integration
 This section describes the expected behaviors for Java EE container APIs other than JAX-RS.
 
-#### javax.security.enterprise.identitystore.IdentityStore.getCallerGroups(CredentialValidationResult)
+#### jakarta.security.enterprise.identitystore.IdentityStore.getCallerGroups(CredentialValidationResult)
 This method should return the set of names found in the "groups" claim in the JWT if it exists, an empty set otherwise.
 
-#### javax.ejb.SessionContext.getCallerPrincipal()
```
sberyozkin, line 682: @Emily-Jiang If some of `jakarta.* -> javax.*` revert updates in TCK are correct then I'd need to update this doc as well to retain `javax.ejb.SessionContext` for ex. Commenting here so that I don't forget

Emily-Jiang (approved):
LGTM!

Emily-Jiang:
@sberyozkin please feel free to merge your PR.

Emily-Jiang:
> @Emily-Jiang @radcortez Something has changed in the `2.1` parent so now `// @formatter off` in `api/src/main/java/org/eclipse/microprofile/jwt/Claims.java` is ignored and it is misformatted when the source is built. Do you know what it can be ?

I checked out your PR to try it out. I think `//@formatter off` is only a tag for telling IDEs to ignore the formatting. Since we are using maven formatter plugin, the annotation was completely ignored.

Emily-Jiang:
You can merge in this change. I spotted more changes to be done to pull in the 3.0-RC4 from MP Config and also the dependency on EJB needs to be updated. In your pom.xml, you need to update the following 

        ```
<version.jakarta.ejb.api>3.2.6</version.jakarta.ejb.api>
<version.microprofile.config>2.0</version.microprofile.config>
```
=>

       ```
 <version.jakarta.ejb.api>4.0.0</version.jakarta.ejb.api>
 <version.microprofile.config>3.0-RC4</version.microprofile.config>
```
You can do this change under the issue #252

spec/src/main/asciidoc/interoperability.asciidoc
rdebusscher: I don't understand @sberyozkin . Do you mean that SessionContext will still be used from the javax namespace? It is also converted to jakarta within Jakarta 9;  https://jakarta.ee/specifications/platform/9/apidocs/jakarta/ejb/sessioncontext

rdebusscher (changes requested):

sberyozkin:
@rdebusscher I've no idea, the compilation was failing with the jakarta package - may be what Emily suggested can help

sberyozkin:
@Emily-Jiang Thanks, but I don't want to release before the compilation succeeds, there is no much point in it :-)

> I checked out your PR to try it out. I think //@formatter off is only a tag for telling IDEs to ignore the formatting. Since we are using maven formatter plugin, the annotation was completely ignored.

It works fine on `master` and likewise it should work with the `2.0.0-RC1` so I'd not like to release before it can be sorted out

thanks

spec/src/main/asciidoc/interoperability.asciidoc
sberyozkin: @rdebusscher The compilation was failing, I'll try @Emily-Jiang's suggestion a bit later on

sberyozkin:
@Emily-Jiang @rdebusscher It compiles with the EJB version update - Rudy if all is good otherwise then please dismiss your change request

sberyozkin:
@Emily-Jiang yes, thanks for noticing those old versions. The only problem is the Claims misformatting but I guess fixing it can wait

Emily-Jiang:
> @Emily-Jiang yes, thanks for noticing those old versions. The only problem is the Claims misformatting but I guess fixing it can wait

The formatter is pretty much personal taste. For me, the refomatted version looks nicer:). I think after merging this PR, we can do a 2.0-RC1. If you find a solution to fix the formatter, we do a RC2 release. In this way, it will not block the implementation to try out the jars.

rdebusscher:
@sberyozkin The  version of JACC is not updated (and still is the javax variant) for the TCK.



we need to replace

```
            <dependency>
                <groupId>jakarta.security.jacc</groupId>
                <artifactId>jakarta.security.jacc-api</artifactId>
                <version>${version.jakarta.security.jacc-api}</version>
                <scope>provided</scope>
                <optional>true</optional>
            </dependency>
```
with

```
<dependency>
  <groupId>jakarta.authorization</groupId>
  <artifactId>jakarta.authorization-api</artifactId>
  <version>2.0.0</version>
  <scope>provided</scope>
  <optional>true</optional>
</dependency>
```

The javax JACC artifact also brings in the javax servlet one which means there are still references to javax.servlet.* in the TCK code.

sberyozkin:
@rdebusscher let me try, thanks

sberyozkin:
@rdebusscher It should be fine now...

Emily-Jiang:
> Hi @Emily-Jiang I agree it is a personal taste but I should say IMHO having 20 or so enums `stringified` is not good, we don't do it all for the users only but for the developers too. If it were 3 enums only then it would not be a big deal, but otherwise it is not cool:
> 
> ```
> // @formatter:off
> public enum Claims {
>     // The base set of required claims that MUST have non-null values in the JsonWebToken
>     iss("Issuer", String.class), sub("Subject", String.class), exp("Expiration Time", Long.class), iat("Issued At Time",
>             Long.class), jti("JWT ID", String.class), upn("MP-JWT specific unique principal name",
>                     String.class), groups("MP-JWT specific groups permission grant",
>                             Set.class), raw_token("MP-JWT specific original bearer token", String.class),
> ```
> 
> And this is just a single fragment of `Claims.java`, many of them are now there.
> I don't think we can offer it to the developers.
> 
> But like I said, I'm fine with doing RC2 once the solution is found but it will need to be found before 2.0

I did not pay much attention to the `value` except to the style. You were right that the current style is lot more readable. I think one solution is to exclude this file from being reformatted. You can add the following to the api/pom.xml to skip that file:
```
<build> 
        <pluginManagement>
            <plugins>
                <plugin>
                    <groupId>net.revelc.code.formatter</groupId>
                    <artifactId>formatter-maven-plugin</artifactId>
                    <version>${version.plugin.formatter}</version>
                    <configuration>
                        <excludes>
                            <exclude>api/src/main/java/org/eclipse/microprofile/jwt/Claims.java</exclude> 
                        </excludes>
                    </configuration>
                </plugin>
            </plugins>
        </pluginManagement>
</build>


```
There might a better solution. Can you take a look to see whether there are no other side effects?

sberyozkin:
Hi @Emily-Jiang I should've guess it myself :-), that should do, I'll try, thanks for trying to help

rdebusscher:
Looks good, I only see some `javax` related code within the `JWT RBAC for MicroProfile Sandbox` but that is not a problem.

Thx

PR #258, opened 2021-09-02
Jakarta EE 9.1 Alignment

sberyozkin:
Aligning it with the master

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <groupId>org.eclipse.microprofile.jwt</groupId>
         <artifactId>microprofile-jwt-auth-parent</artifactId>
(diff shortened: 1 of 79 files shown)


rdebusscher:
I only have a problem when building the project (Maven 3.6.3) with the Maven-resource-plugin 3.2.0 (coming from MicroProfile parent 2.1) that results in an exception.

` A required class was missing while executing org.apache.maven.plugins:maven-resources-plugin:3.2.0:copy-resources: org/apache/maven/shared/filtering/MavenFilteringException
`

using version 3.1.0, all is ok.

rdebusscher (approved):
LGTM

sberyozkin:
Thanks @rdebusscher, I'm on Maven 3.8., so did not see a problem

PR #262, opened 2021-09-06
Build TCK Tests JAR

### LINKED ISSUE #261: microprofile-jwt-auth-tck tests jar is no longer built
microprofile-jwt-auth-tck tests jar is no longer built, after an update to MP Parent 1.3 (so the problem started happening before upgrading it further to 2.1).
How to reproduce: 
- checkout the master branch at the commit preceding an update to the 1.3 parent and confirm the tests jar is built
- checkout the master branch at the commit updating it to the 1.3 parent and confirm no tests jar is built

Hi Roberto, @radcortez, I looked at `tck/pom.xml` changes related to the 1.3 parent update and did not see anything that could've affected it - the problem may be in the parent pom somewhere...

radcortez:
- Fixes #261

### DIFF
--- a/tck/pom.xml
+++ b/tck/pom.xml
@@ -89,5 +89,19 @@
                 </plugin>
             </plugins>
         </pluginManagement>
+
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
     </build>
 </project>


sberyozkin:
Thanks @radcortez :-)

sberyozkin (approved):

Emily-Jiang:
By the way, I noticed this repo is not the only containing the test folder. I'll do a fix in the microprofile-parent. The other issue is that the source jar and javadoc jar are not built either for this test folder, which are not good. I'll do a fix on the parent for all of the three issues. This commit can be undone when the new microprofile-parent 2.2 is released. Plan to get it done today.

Emily-Jiang:
https://github.com/eclipse/microprofile-parent/pull/52 @radcortez @sberyozkin please take a look to see whether it is ok. I tried and it does build all three jars for the test dir.

PR #265, opened 2021-09-07
add javadoc and source for test-jar

Emily-Jiang:
Fixes: #261 (additional fixes)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ test-output
 
 # Ignore a release.conf for perform_release/* script usage
 release.conf
+.vscode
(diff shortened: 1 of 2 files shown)


Emily-Jiang:
@radcortez @sberyozkin I have added the source and javadoc for the test jar since Roberto prefers to fix here instead of the parent pom. I am ok to fix here so that the test effort is smaller. We should encourage to use src/main for tcks instead of src/test.

sberyozkin (approved):
@Emily-Jiang Thanks Emily, I recall now seeing some of it in the earlier versions of this pom

PR #271, opened 2021-11-03
Bump MP Config version to 3.0

sberyozkin:
The build will fail as MP Config is not in Central yet


sberyozkin:
Going ahead to stage 2.0

PR #291, opened 2022-05-14
Move publickey.algorithm section

dblevins:
The `publickey.location` section needs to be followed by these subsections to specify how people can supply the value of `publickey.location`.

- Relative Path
- `file:` URL Scheme
- `http:` URL Scheme
- Other URL Schemes


sberyozkin (approved):
Thanks @dblevins

PR #292, opened 2022-05-15
Add support for RSA-OAEP-256 key management algorithm

### LINKED ISSUE #289: Add `mp.jwt.decrypt.key.algorithm` property
Add `mp.jwt.decrypt.key.location` property to let users choose the algorithm which should be used to decrypt the content encryption key. `RSA-OAEP` will remain a default value for `2.1` to maintain the backward compatibility, but `RSA-OAEP-256` will also be required to be supported (might become a new default value in one of the next major releases, see  #288), support for other key management algorithms will be optional.

sberyozkin:
Fixes #289.

See #289 for all the details and specifically [this comment](https://github.com/eclipse/microprofile-jwt-auth/issues/289#issuecomment-1122205115).
In summary, an `mp.jwt.decrypt.key.algorithm` is introduced to prepare a transition to `RSA-OAEP-256` becoming a default algorithm used to decrypt the content encryption key. `RSA-OAEP` remains a default algorithm in 2.1. `RSA-OAEP-256` will become a new default in the next major release. `mp.jwt.decrypt.key.algorithm` might become a list property too.

CC @teddyjtorres @rdebusscher

### DIFF
--- a/spec/src/main/asciidoc/configuration.asciidoc
+++ b/spec/src/main/asciidoc/configuration.asciidoc
@@ -381,7 +381,7 @@ Note that two types of keys are required to implement a JWE encryption scheme:
 * Content encryption key - typically a generated secret key which is used to encrypt a plaintext such as a JSON representation of the token claims.
 * Key management key - public RSA key which is used to encrypt a content encryption key. `mp.jwt.decrypt.key.location` must point to a private RSA key matching this key.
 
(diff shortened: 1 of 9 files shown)


rdebusscher (approved):
I still don't like the fact that we are so restrictive that you can't specify multiple algorithms. ( this has nothing to do with being more secure)
We could easily add the config key as a list and as default value only define RSA-OAEP.

sberyozkin:
@rdebusscher Hi Rudy, we can definitely do it in a follow up major release if you or someone else will find it practically necessary (and similarly for the existing verification algorithm property).
I just think right now that the proper solution is a multi-tenancy configuration. Situation where we have a list of supported algorithms, while only supporting a single issuer, does appear to be not very realistic for a given endpoint.
thanks

PR #306, opened 2022-10-21
If "mp.jwt.decryption.algorithm" is not set both RSA-OAEP and RSA-OAEP-256 must be accepted

dblevins:
Took a stab at adding the tests we discussed.  I called it `DecryptKeyAlgorithmDefaultTest` as I thought we could maybe create a pattern of having tests named explicitly after the configuration property.  So in the future we might have `TokenAgeTest`, `ClockSkewTest`, etc.  If we define specific behavior for when they're not set we'd have `TokenAgeDefaultTest`, `ClockSkewDefaultTest`, etc. 

Just brainstorming.  Happy to rename things.


sberyozkin (approved):
Thanks @dblevins, LGTM

Issue #310, opened 2022-11-04
API 2.1 excludes EE 10 versions in Manifest


arjantijms:
The Manifest of the latest 2.1 API (RC6) contains the following import line:

```
Import-Package: jakarta.enterprise.util;version="[3.0,4)",jakarta.inje
 ct;version="[2.0,3)",jakarta.json;version="[2.0,3)"
```

This will fail on EE 10 OSGi environments such as GlassFish 7:

```
org.glassfish.embeddable.GlassFishException: Unable to resolve
    org.glassfish.main.webservices.connector [217]
    missing requirement
        &(package = org.glassfish.web.deployment.descriptor) (version >= 7.0.0) (!(version >= 8.0.0))
        caused by:
            Unable to resolve
                org.glassfish.main.web.glue [123]
                missing requirement
                    &(package = org.omnifaces.jwt.servlet) (version >= 2.0.0) (!(version >= 3.0.0))
                    caused by:
                        Unable to resolve
                            org.omnifaces.jwt [195]
                            missing requirement
                                &(package = org.eclipse.microprofile.jwt) (version >= 2.0.0) (!(version >= 3.0.0))
                                caused by:
                                    Unable to resolve
                                        org.eclipse.microprofile.jwt [117]
                                        missing requirement
                                            &(package = jakarta.enterprise.util) (version >= 3.0.0) (!(version >= 4.0.0)))]]]
```

Emily-Jiang:
> @Emily-Jiang Hi, what is the right fix here such that it is consistent with other MP specs ? I believe OpenLiberty also has some OSGI constraints ?

Since MP JWT 2.1 works with Jakarta EE 9.1 and it uses the "Compile low, run high", this spec primarily works with EE 9.1 but also works with Jakarta EE 10 Core Profile. In Open Liberty, we did manifest rewrite to allow the bundlised version working with EE 10. I think we can fix this in the bnd.bnd to make the version range inculde EE 10 range if necessary.

radcortez:
I agree that it should be open to allow "compile low, run high", but it doesn't seem consistent with the discussion on the ML when @dblevins brought this up.

If we enforce a minimum on EE 10 at the platform level, then all platform specifications should also enforce the minimum requirements.

Emily-Jiang:
> I agree that it should be open to allow "compile low, run high", but it doesn't seem consistent with the discussion on the ML when @dblevins brought this up.
> 
[...]

PR #311, opened 2022-11-07
Update bnd.bnd to allow this bundle manifest to work with EE 10

Emily-Jiang:
Fixes: #310


sberyozkin:
Thanks Emily, looks like we are OK here, as this PR only allows MP JWT 2.1 to run in EE10 containers, but I'd like to wait a bit in case Roberto @radcortez or David @dblevins have some concerns

sberyozkin (approved):
Let me merge it now, if anything relates comes up before the final release we can fix it but hopefully it won't be needed

Emily-Jiang:
@sberyozkin we need to restage the final release. Do you want to do it or you want me to do it?

PR #332, opened 2024-04-11
Recommend getting the sub claim to guarantee uniqueness in the JsonWebToken.getName JavaDocs

sberyozkin:
My colleague spotted the problem, `preferred_username` is not unique, while JsonWebToken#getName talks about uniqueness.

This PR gently warns users that if they really need a unique principal identifier they should get a subject claim.
Ideally we'd remove any explicit mentioning of preferred_username - but it is hard to do right without possibly breaking a lot of user applications. May be in some basic cases it can be unique,  but definitely not in the OIDC space

CC @ayoho

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
+++ b/api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
@@ -31,8 +31,11 @@ import java.util.Set;
  */
 public interface JsonWebToken extends Principal {
     /**
-     * Returns the unique name of this principal. This either comes from the upn claim, or if that is missing, the
-     * preferred_username claim. Note that for guaranteed interoperability a upn claim should be used.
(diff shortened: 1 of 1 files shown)


starksm64 (approved):

api/src/main/java/org/eclipse/microprofile/jwt/JsonWebToken.java
```diff
@@ -31,8 +31,11 @@
  */
 public interface JsonWebToken extends Principal {
     /**
-     * Returns the unique name of this principal. This either comes from the upn claim, or if that is missing, the
-     * preferred_username claim. Note that for guaranteed interoperability a upn claim should be used.
+     * Returns the unique name of this principal. The upn claim is checked first, the preferred_username claim is
+     * checked next, and finally, the sub claim is checked. Note that for guaranteed interoperability a upn claim should
```
ayoho, line 35: We should consider adding a test in the TCK for this. Unless there's already one in there that I missed while looking just now.

PR #333, opened 2024-06-20
Require that both RS256 and ES256 must be supported if the signature algorithm is not configured

sberyozkin:
This PR is meant to align a situation with a non configured signature algorithm is managed with how it is done for the unconfigured decryption algorithm.

If a decryption algorithm is not configured, both `RSA-OAEP` and `RSA-OAEP-256` encryped tokens must be accepted.

Now, similarly, if a signature algorithm is not configured, both `RS256` and `ES256` signed tokens must be accepted.

The end goal is to let an MP JWT Bridge [JwtAuthenticationMechanismDefinition](https://github.com/eclipse/microprofile-jwt-bridge/blob/main/api/src/main/java/org/eclipse/microprofile/jwt/bridge/authentication/mechanism/JwtAuthenticationMechanismDefinition.java) offer a consistent List representation for both signature and encryption properties

CC @ayoho @jimmy1wu

### DIFF
--- a/spec/src/main/asciidoc/configuration.asciidoc
+++ b/spec/src/main/asciidoc/configuration.asciidoc
@@ -356,7 +356,8 @@ return Public Key text in one of the supported formats.
 #### `mp.jwt.verify.publickey.algorithm`
 
 The `mp.jwt.verify.publickey.algorithm` configuration property allows for specifying which Public Key Signature Algorithm
-is supported by the MP JWT endpoint. This property can be be set to either `RS256` or `ES256`. Default value is `RS256`.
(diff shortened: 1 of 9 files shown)


sberyozkin:
There is a real lot of assertions in TCK tests expecting RS256 by default

sberyozkin:
That said, the spec does not require anywhere that the signature algorithm must be injected as a string and not as a set of strings.

ayoho (approved):

sberyozkin:
I've now verified this PR with the smallrye-jwt mpjwt2.2 branch, merging

PR #336, opened 2024-09-26
Bump MP parent version to 2.8

sberyozkin:


sberyozkin:
I'll merge once the build passes, @ayoho, FYI

sberyozkin:
@Emily-Jiang FYI, as well, I had to align the build.xml with that of the MP Config one

PR #338, opened 2025-02-13
Remove DTO declarations from TCK test suite files

### LINKED ISSUE #335: MP JWT TCK test run fails with Maven 3.9.3 or later
The failure:

```
[ERROR] Suite file .../testsuite/tck/target/tck-suite/suites/tck-base-suite.xml is not a valid file
```

@ayoho Hey Adam, can you verify independently please ? Please run an MP JWT TCK test for OpenLiberty with Maven 3.9.3 or later. I think we may need to fix the way TCK jar is assembled

sberyozkin:
Fixes #335


sberyozkin:
I'm going to merge it without waiting for an approval as it is impossible to work with the TCK test suite otherwise, having to switch between multiple Maven versions

